PR #12, opened 2023-03-30
Fetch Passport Scores

boudra:
Resolves https://github.com/gitcoinco/grants-stack/issues/1295
Related PR: https://github.com/gitcoinco/chainsauce/pull/3

- Fetch passport scores so we can use them in the matching calculator, the whole process takes around 4 minutes, read this for more context: https://discord.com/channels/562828676480237578/1090181976159420416/1090181979632320592
- Implements an init script for the Fly VM that first indexes everything and then launches the HTTP server, to make deployments stateless and with zero downtime. It also restarts failed indexers to avoid crashing the VM.

The passport data is fetched through a CLI script that can run once or fetch continuously with the `--follow` option. 

```
npm run passport # run once
npm run passport -- --follow # continuously
```

The worker writes two files:

- `passport_scores.json` all the passport scores
- `passport_valid_addresses.json` all the addresses that pass the threshold check

The passport fetching logic was copied over from the API.


hmrtn (approved):
lgtm and `init.sh` works as expected!

PR #15, opened 2023-03-31
Fetch prices in one hour chunks indexed by block number

boudra:
closes https://github.com/gitcoinco/grants-stack/issues/1309

Research:

Fetching prices in the reducer is hard because we need a timestamp, and to get a timestamp we have to call `getblock` through the JSON-RPC, even when cached or when chunked into one hour or even one day, it takes too much time to index all the votes.

Instead, this PR takes a different approach, it indexes the prices separately into another file which the reducer can use. The prices can be queried by timestamp and block number, so they can be used by the indexer and the data team.

Another challenge has been getting the closest block number from the timestamp. Here's what I have tried;

- Doing a binary search (https://medium.com/@hanyi.lol/how-to-get-a-block-number-by-timestamp-fefde4c69162): it's OK but you have to make a couple of requests to get the block number, this becomes unfeasible when getting more than 1000 prices.
- Estimating the block number and then walking to the closest block: I wrote my own inspired form this code (https://github.com/ethfinex/efx-trustless-vol/blob/master/src/lib/getBlockByTime.js), this was the fastest but you still had to make up to 3 requests to get the block number.
- I also tried this library, it wasn't very fast: https://github.com/monosux/ethereum-block-by-date
- DefiLlama has an endpoint to do this but as soon as I made more than a couple requests the server crashed: https://defillama.com/docs/api
- Etherscan have an endpoint to do this but we would have to get an API key for all their scanners separately for each network

**In the end I realised that calculating the average of blocks per second by checking how much time hast passed between the last 1000th block and the latest one and using that to estimate the block number gives you a pretty close estimate with only minutes of margin error, which should be fine for our use case.**

The script to fetch all the prices from Jan 1st, 2023 to now takes a few seconds to run, it's important that it's fast because the blockchain indexer needs it to be finished before running, when called with `---follow` it will fetch new prices every hour, but in general I am trying to keep the indexer fast and local so we can run it anywhere and make deployments easier.


nfrgosselin:
Defer to engineering review

Dockerfile
```diff
@@ -17,4 +17,7 @@ EXPOSE 8080
 
 COPY init.sh ./
 
+# Cache indexer in the image
+# RUN ./init.sh
```
hmrtn, line 21: If this is not, let's remove it

hmrtn (approved):

PR #20, opened 2023-04-11
Implement new beta rounds contracts

boudra:
This handles the new / changed events:

- ApplicationStatusesUpdated
- Voted

Introduces the concept of Applications which allows us to track different Applications for the same Projects.


src/handleEvent.ts
```diff
@@ -181,18 +279,68 @@ async function handleEvent(indexer: Indexer, event: Event) {
...
-            status: projectApp.status,
+            status: projectApp.status ?? application.status,
             payoutAddress: projectApp.payoutAddress,
           }));
       }
       break;
     }
 
+    case "ApplicationStatusesUpdated": {
+      const bitmap = new StatusesBitmap(BigInt(128), BigInt(2));
```
gravityblast, line 290: @boudra  the first param (width) should be `256` which is the maximum in solidity. so that internally it does 
`this._itemsPerRow = 256 / 2; // = 128 `

boudra: ah cool I understood it as number of statuses, it makes sense

gravityblast (approved):
LGTM

PR #28, opened 2023-04-14
add: min donation config

hmrtn:
test with something like: `curl "localhost:3000/chains/1/rounds/0x1b165fE4DA6bC58AB8370DDC763d367D29F50Ef0/matches?minimumAmount=0"` and altering `minimumAmount`. If `minimumAmount` is not provided then it will attempt to use the `round.minimumAmount` field in the round of the `rounds.json`. If both above fail, then it will use `minimumAmount=0`.

fixes https://github.com/gitcoinco/grants-stack/issues/1312

side note: we need tests - spent some time getting them running but decided that was out of scope for this pr

### DIFF
--- a/src/calculator.ts
+++ b/src/calculator.ts
@@ -17,17 +17,20 @@ export default class Calculator {
   private chainId: string;
   private roundId: string;
   private matchAmount: number;
+  private minDonationThreshold: number;
 
   constructor(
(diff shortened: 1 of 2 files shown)


boudra (approved):

nfrgosselin:
Defer to eng review on this

PR #29, opened 2023-04-15
Passport integration

hmrtn:
adds passport integration: 
`curl "localhost:3000/chains/1/rounds/0x1b165fE4DA6bC58AB8370DDC763d367D29F50Ef0/matches?minimumAmount=0&passportThreshold=0&passport=true"`

If there is no passport score available then it sets the threshold to 0.


src/calculator.ts
```diff
@@ -1,54 +1,86 @@
...
+          contributor: raw.voter,
+          recipient: raw.projectId,
+          amount: raw.amountUSD,
+        })
     );
 
+    contributions = contributions.filter((c: Contribution) => {
+      const addressData = passportScores.find((ps: any) => ps.address === c.contributor);
+      const hasValidEvidence = addressData?.evidence?.success;
+      const isValidAddress = validAddresses.includes(c.contributor);
```
boudra, line 60: The `passport_valid_addresses.json` is just a list of the addresses that have valid evidence, which is derived from the scores. so this check is not needed, just checking the evidence should be enough.

src/calculator.ts
```diff
@@ -1,54 +1,86 @@
 import fs from "fs";
 import { linearQF, Contribution, Calculation } from "pluralistic";
+import {AugmentedResult, CalculatorOptions, RawContribution} from "./types.js";
```
boudra, line 3: I think we should keep these types inside this file for now. If the responsibility of this file ever becomes unmanageable we can think about how to split it up in features, but not a generic `types.ts` file.

boudra: Either way we need to start thinking on how to organize this codebase and split up the different modules in directories a little bit, for example of the top of my head: CLI, prices, blockchain indexer, calculator, http server.

src/calculator.ts
```diff
@@ -1,54 +1,86 @@
...
-  amountUSD: number;
-}
 
 export default class Calculator {
   private baseDataPath: string;
   private chainId: string;
   private roundId: string;
   private matchAmount: number;
+  private minimumAmount: number | undefined;
+  private passport: boolean | undefined;
```
boudra, line 12: Could we have just a `passportThreshold?: number` here to accomplish the same thing, isn't the boolean redundant?

src/calculator.ts
```diff
@@ -1,54 +1,86 @@
...
 
+    contributions = contributions.filter((c: Contribution) => {
+      const addressData = passportScores.find((ps: any) => ps.address === c.contributor);
+      const hasValidEvidence = addressData?.evidence?.success;
+      const isValidAddress = validAddresses.includes(c.contributor);
+
+      if (this.passport) {
+        if (typeof this.passportThreshold !== 'undefined') {
+          return (
+              c.amount >= (this.minimumAmount ?? currentRound.minimumAmount ?? 0) &&
```
boudra, line 65: I see `this.minimumAmount ?? currentRound.minimumAmount` happens 4 times in this function, what if we do it in the constructor of the class? we store the right minimumAmount to use, or keep it like this but make a getter of some kind.

src/calculator.ts
hmrtn: Agreed - this was my way of saying we should be thinking about this.  We can spike it down the line.

src/calculator.ts
hmrtn: I thought about this - and I think it could be necessary.

For example, if a user wanted to use passport but with the set values. 
The `passportThreshold` is meant to overwrite the default value, if any. 

e.g. 
To use passport with the configured passport threshold, they set `?passport=true`. 
If they want to overwrite the threshold - then they use `?passport=true&passportThreshold=x`


I suppose we could narrow this down to a singular variable - and just parse whether its a string or number, such as `?passportThreshold=x | ?passportThreshold=default`

src/calculator.ts
hmrtn: `calculate` method has been refactored to reduce redundancy while preserving the original behavior. I moved the calculation of `minAmount` is moved outside the filter function, and a separate function `isEligible` is introduced to determine the eligibility of a contribution based on the passport and passport threshold params.

src/calculator.ts
boudra: that makes sense!

boudra (approved):

PR #31, opened 2023-04-20
Matching Calculations v2

boudra:
- Load match amount from round contract and convert it to USD
- Use applications instead of projects to calculate matching amounts
- Add more data needed on frontend for the matches
- Fix passport eligibility taking minutes to finish, it was O(N^2), it is now instant


hmrtn (approved):
wonderful optimization catch and fix @boudra ! 

Approved.

PR #34, opened 2023-04-25
avoid failing if application is not found

gravityblast:


boudra (commented):
application here should never be null no? so maybe instead of `?` we can just check and throw an error or something

boudra (approved):

gravityblast:
@boudra you are right! I pushed a new commit that should make it work for old and new applications/votes, wdyt?

PR #47, opened 2023-04-27
Add from block option

boudra:
This is useful for indexing only the last rounds and have a faster feedback loop


vacekj:
Build bcs of lint but otherwise LGTM and very useful option 🚀

PR #49, opened 2023-04-27
feat: add support for vote coefficients

vacekj:
This adds support for creating and downloading the vote coefficients as a CSV, which is a requirement for https://github.com/gitcoinco/grants-stack/issues/1475

### DIFF
--- a/src/test/server.test.ts
+++ b/src/test/server.test.ts
@@ -1,0 +1,144 @@
+import { describe, test, expect } from "vitest";
+import fs from "fs";
+import path from "path";
+import request from "supertest";
+import { app, calculatorConfig } from "../server.js";
+import { FileNotFoundError } from "../calculator.js";
+
+const loadFixture = (name: string) => {
+  const p = path.resolve(__dirname, "fixtures", `${name}.json`);
+  const data = fs.readFileSync(p, { encoding: "utf8", flag: "r" });
+  return data;
+};
+
+class TestDataProvider {
+  routes: { [path: string]: string };
+
+  // eslint-disable-next-line
+  constructor(routes: { [path: string]: string | any }) {
+    this.routes = routes;
+  }
+
+  loadFile(description: string, path: string) {
+    const fixture = this.routes[path];
+    if (fixture === undefined) {
+      throw new FileNotFoundError(description);
+    }
+
+    if (typeof fixture !== "string") {
+      return fixture;
+    }
+
+    return JSON.parse(loadFixture(fixture));
+  }
+}
+
+describe("server", () => {
+  describe("/matches", () => {
+    describe("resources not found", () => {
+      test("should render 404 if round is not present in rounds.json", () =>
+        new Promise((done) => {
+          calculatorConfig.dataProvider = new TestDataProvider({
+            "1/rounds/0x1234/votes.json": "votes",
+            "1/rounds/0x1234/applications.json": "applications",
+            "1/rounds.json": [], // empty file so the round won't be found
+            "passport_scores.json": "passport_scores",
+          });
+
+          request(app).get("/chains/1/rounds/0x1234/matches").expect(404, done);
+        }));
+
+      test("should render 404 if rounds file doesn't exist", () =>
+        new Promise((done) => {
+          calculatorConfig.dataProvider = new TestDataProvider({
+            "1/rounds/0x1234/votes.json": "votes",
+            "1/rounds/0x1234/applications.json": "applications",
+            "1/rounds.json": undefined,
+            "passport_scores.json": "passport_scores",
+          });
+
+          request(app).get("/chains/1/rounds/0x1234/matches").expect(404, done);
+        }));
+
+      test("should render 404 if votes file doesn't exist", () =>
+        new Promise((done) => {
+          calculatorConfig.dataProvider = new TestDataProvider({
+            "1/rounds/0x1234/votes.json": undefined,
+            "1/rounds/0x1234/applications.json": "applications",
+            "1/rounds.json": "rounds",
+            "passport_scores.json": "passport_scores",
+          });
+
+          request(app).get("/chains/1/rounds/0x1234/matches").expect(404, done);
+        }));
+
+      test("should render 404 if applications file doesn't exist", () =>
+        new Promise((done) => {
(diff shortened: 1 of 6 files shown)


vacekj:
@boudra @gravityblast one thing that's weird is that when the coefficient is 0, it's serialized az `0E-9` in the csv. any idea why this might be?

src/server.ts
```diff
@@ -86,16 +89,72 @@ app.get("/data/:chainId/rounds/:roundId/applications.csv", async (req, res) => {
...
-      ...answers,
+      ...answers
     ]);
   }
 
   res.setHeader("content-type", "text/csv");
   res.send(csv.getHeaderString() + csv.stringifyRecords(records));
 });
 
+app.get("/data/:chainId/rounds/:roundId/votecoefficients.csv", async (req, res) => {
```
boudra, line 100: vote and coefficients are two words, we should split them following the convention, like `passport_scores.json`

```suggestion
app.get("/data/:chainId/rounds/:roundId/vote_coefficients.csv", async (req, res) => {
```

src/server.ts
```diff
@@ -86,16 +89,72 @@ app.get("/data/:chainId/rounds/:roundId/applications.csv", async (req, res) => {
...
+      ...passport_score
+    } = passportScores.find(score => score.address.toLowerCase() === vote.voter.toLowerCase()) ?? {};
+    return {
+      ...vote,
+      ...passport_score,
+      ...evidence
+    };
+  });
+
+  const records = voteCoefficients.map((voteCoefficient) => Object.values(voteCoefficient));
```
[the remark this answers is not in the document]
boudra: To add to this, keep in mind that the votes file could potentially have lots of votes, we should probably do the join with passport and convert to CSV record in one pass, instead of walking a big array twice.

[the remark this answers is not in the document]
vacekj: Good catch, I was thinking this might be an issue. Will fix.

src/server.ts
```diff
@@ -86,16 +89,72 @@ app.get("/data/:chainId/rounds/:roundId/applications.csv", async (req, res) => {
...
+  /*TODO: type this once we have unified types*/
+  const passportScores: any[] = JSON.parse(data);
+
+  const voteCoefficients = votes.map((vote) => {
+    const {
+      address,
+      evidence,
+      error,
+      ...passport_score
+    } = passportScores.find(score => score.address.toLowerCase() === vote.voter.toLowerCase()) ?? {};
```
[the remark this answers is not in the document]
hmrtn: you're right the current implementation's time complexity is O(m * n), which would definitely would have a significant performance impact. A more efficient approach would be to preprocess the passport scores into an object with constant-time lookups.

src/server.ts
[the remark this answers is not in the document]
hmrtn: fixed in latest push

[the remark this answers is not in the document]
boudra: we're still looping through the array twice, once to join with passport and the second time to convert it to a CSV row, we should do this in one pass

src/http/index.ts
```diff
@@ -87,16 +90,91 @@ app.get("/data/:chainId/rounds/:roundId/applications.csv", async (req, res) => {
...
+  };
+
+  const passportScoresMap = processPassportScores(passportScores);
+
+  const voteCoefficients: any[] = votes.map((vote: any) => {
+    const voter = vote.voter.toLowerCase();
+    return { ...vote, ...passportScoresMap[voter] ?? {} };
+  });
+
+  const records = voteCoefficients.map((voteCoefficient) => [
```
boudra, line 132: We're still unnecessarily looping through the array twice, once to join with passport and the second time to convert it to a CSV row, we should do this in one pass

gravityblast: @boudra @hmrtn @vacekj 

Maybe we can improve it even more if instead of loading `passport_scores.json`, we load `passport_valid_addresses.json` directly, so we don't need to loop and check the coefficients, they are already the smaller subset of valid addresses.

We can modify `passport/index.ts` to lowercase the addresses, so we don't need to do it again here. 

We can modify `passport/index.ts` to store an object instead of an array, so we load it and it's already in the final shape we want.

We can try with different data structures to see which one has the fastest lookup among Object, Set, and Map.
I benchmarked with a set of 9 millions strings of 42 characters (like eth addresses), and the average lookups with those data structures are:

```
avgSet 0.0011645 ms
avgMap 0.00043525 ms
avgObj 0.000993375 ms
```

wdyt?

hmrtn: @gravityblast The `passport_valid_addresses.json` approach is a good idea, and from the quick test it appears Map is the fastest. Alternatively, we could cache the passportScoresMap to avoid recalculating it on every request - if we expect it to not be updated too often. 

@boudra It will still be linear but this is true - we can further optimize this to a single map!

boudra: I like the idea of writing an object directly! Maybe we could even just store the threshold as the value instead of all that extra data now that we know it's what we need.

But for me this is now good enough to merge if we want to do this later, together with the calculator code that also accesses this data. 

My intention here was only to make sure we're not pushing very resource intensive code unnecessarily, not really optimizing.

gravityblast: yes let's try my optimization in another PR!

but we need to fix the loop in this PR to avoid looping more than once.

hmrtn: this was resolved, as far as I could tell, in the latest commit!

boudra: yeah it looks good! @hmrtn

boudra (approved):

src/test/server.test.ts
```diff
@@ -0,0 +1,116 @@
+import { describe, test, expect } from "vitest";
```
boudra, line 1: @hmrtn @vacekj is there any reason we need this extra test file? it looks like it's a duplicate of `test/http/app.test.ts` it has the exact same tests rewritten in slightly different ways

vacekj: I don't think so, it might just be a duplicate from resolving merge conflicts. Can delete in a further PR.

boudra: Awesome I am deleting it in this PR since it was breaking my build: https://github.com/gitcoinco/allo-indexer/pull/55

PR #56, opened 2023-05-02
add: csv matching endpoint

hmrtn:
this is from the deprecated branch, hence contains the older format

- [ ] update to new matchHanlder-like handling...

### DIFF
--- a/src/http/app.ts
+++ b/src/http/app.ts
@@ -322,47 +322,20 @@ app.get("/chains/:chainId/rounds/:roundId/matches.csv", async (req, res) => {
       ],
     });
 
-    const records = [];
-
-    for (const match of matches) {
-      records.push([
-        match.matched,
-        match.contributionsCount,
-        match.sumOfSqrt,
-        match.totalReceived,
-        match.projectId,
-        match.applicationId,
-        match.payoutAddress,
-        match.projectName
-      ]);
-    }
+    const records = matches.map((match: any) => ([
+      match.matched,
+      match.contributionsCount,
+      match.sumOfSqrt,
+      match.totalReceived,
+      match.projectId,
+      match.applicationId,
+      match.payoutAddress,
+      match.projectName
+    ]));
 
     res.setHeader("content-type", "text/csv");
     res.send(csv.getHeaderString() + csv.stringifyRecords(records));
   } catch (e) {
-    if (e instanceof FileNotFoundError) {
-      res.statusCode = 404;
-      res.send({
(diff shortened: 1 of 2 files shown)


hmrtn:
Reopened this in reference to: https://github.com/gitcoinco/allo-indexer-client/pull/12 in regards to the ticket: https://github.com/gitcoinco/grants-stack/issues/1475

src/http/app.ts
```diff
@@ -283,3 +283,59 @@ app.post(
     matchesHandler(req, res, 201, true);
   }
 );
+
+app.get("/chains/:chainId/rounds/:roundId/matches.csv", async (req, res) => {
```
boudra, line 287: We shouldn't duplicate the matches handler since it's exactly the same as the other one, the only difference is the output format. We can extract it to it's own function and then from the return value, encode the data in JSON or CSV.  You can also create a CSV from an array of objects: https://www.npmjs.com/package/csv-writer#createobjectcsvwriterparams

src/test/server.test.ts
```diff
@@ -67,6 +67,40 @@ describe("server", () => {
         const res = await request(app).get("/chains/1/rounds/0x1234/matches");
         expect(res.status).toBe(404);
       });
+      
+    });
+
+    describe("matches csv", () => {
```
boudra, line 73: This test should be in `src/test/http/app.test.ts`

src/http/app.ts
vacekj: As I've written before: we don't need this endpoint. We already have the match data on the frontend, and the code to transform it from json to csv is already implemented, working and tested. Has been for more than a week. I really don't know the reason for reopening this.

boudra: @vacekj that makes a lot of sense, sorry I missed your previous message on this! then to conclude, unless we can think of another reason we need this endpoint in the server I would go with your solution for the reasons you outlined. @gravityblast is this ok with you?

gravityblast: yes we decided in the past that this PR was a duplicate of the other where we create the CSV in the fronend, we can close it. @vacekj did we already merge the other one?

boudra: not yet, it's this one, we should review it: https://github.com/gitcoinco/grants-stack/pull/1672

vacekj: It's part of the new PR i raised for #1475

boudra: closing this then thank you for your input everyone

PR #57, opened 2023-05-03
add: ignore saturation config param

hmrtn:


src/calculator/index.ts
```diff
@@ -108,13 +109,14 @@ type RawRound = {
   id: string;
 };
 
-export default class Calculator {
+export default class  {
```
gravityblast, line 112: can we put back the name just to have it more explicit?

src/calculator/index.ts
```diff
@@ -205,7 +209,7 @@ export default class Calculator {
 
     const results = linearQF(contributions, round.matchAmountUSD, {
       minimumAmount: this.minimumAmount ?? round.minimumAmount ?? 0,
-      ignoreSaturation: true,
+      ignoreSaturation: this.ignoreSaturation ?? true,
```
gravityblast, line 212: let's put false as default, so it's not scaled up unless the operator wants it

gravityblast (changes requested):

src/calculator/index.ts
hmrtn: 🐙 oops accidentally deleted that! yup!

gravityblast (approved):

DanieleSalatti (approved):

PR #58, opened 2023-05-04
add coefficient value to the already existing coefficient column in the CSV export

gravityblast:


boudra (approved):

src/http/app.ts
```diff
@@ -153,15 +155,21 @@ app.get("/data/:chainId/rounds/:roundId/vote_coefficients.csv", async (req, res)
...
     const passportScoresMap = processPassportScores(passportScores);
 
     const records = votes.map((vote: any) => {
       const voter = vote.voter.toLowerCase();
-      const combinedVote = { ...vote, ...passportScoresMap[voter] ?? {} };
+      const score = passportScoresMap[voter];
+      const coefficient = score !== undefined ? 1 : 0;
+      const combinedVote = {
+      ...vote,
+      ...(score ?? {}),
```
boudra, line 170: score here is `bool` or `undefined` so it cannot be spread

src/http/app.ts
```diff
@@ -153,15 +155,21 @@ app.get("/data/:chainId/rounds/:roundId/vote_coefficients.csv", async (req, res)
   try {
     const [votes, data] = await Promise.all([
       db.collection(`rounds/${roundId}/votes`).all(),
-      fs.promises.readFile("./data/passport_scores.json", { encoding: "utf8", flag: "r" }),
+      fs.promises.readFile("./data/passport_valid_addresses.json", { encoding: "utf8", flag: "r" }),
```
boudra, line 158: do you think it's useful to return the score as well in the CSV in addition to the coefficient being 1 or 0?

boudra (changes requested):

src/http/app.ts
gravityblast: right I left the same as before, but from the other comment I realized it's better to return all the scores as you said. I'll change it now

src/http/app.ts
```diff
@@ -156,12 +186,18 @@ app.get("/data/:chainId/rounds/:roundId/vote_coefficients.csv", async (req, res)
...
 
     const records = votes.map((vote: any) => {
       const voter = vote.voter.toLowerCase();
-      const combinedVote = { ...vote, ...passportScoresMap[voter] ?? {} };
+      const score = passportScoresMap[voter];
+      const coefficient = score !== undefined ? 1 : 0;
+      const combinedVote = {
+      ...vote,
+      ...(score ?? {}),
+      coefficient,
```
boudra, line 199: score already has a coefficient right? this would be overriding it

boudra: but in the case that the score is not found in the passport file we want to show 0 as coefficient

gravityblast: yes I'll push in a second because I changed only the firs part

gravityblast: I changed it back to draft

boudra (approved):

src/http/app.ts
```diff
@@ -137,13 +137,43 @@ app.get("/data/:chainId/rounds/:roundId/applications.csv", async (req, res) => {
...
-const processPassportScores = (scores: any[]) => {
+type BasePassportScore = {
+  address: string;
+  score: string | null;
+  status: string;
+  last_score_timestamp: string;
+};
+
+type FullPassportScore = BasePassportScore & {
+  evidence: {
```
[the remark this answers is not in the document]
gravityblast: @hmrtn good point let me try

hmrtn (approved):

PR #59, opened 2023-05-04
BigInt conversion for minimum amount and matching cap

boudra:
Those values could potentially contain decimals, so we first convert them to BigInt and then to token amount.

For the min amount we keep 6 decimals, and two decimals for the matching cap percentage.

### DIFF
--- a/src/calculator/index.ts
+++ b/src/calculator/index.ts
@@ -186,9 +186,9 @@ export default class Calculator {
       round.metadata?.quadraticFundingConfig?.sybilDefense ??
       false;
 
-    // 1. convert threshold amount to 6 tokenDecimals
+    // 1. convert threshold amount to 6 decimals
     // 2. truncate rest of decimals to bigint
-    // 3. convert decimals to tokendecimals
+    // 3. convert decimals to token decimals
     // 4. remove initial 6 decimals
     const minimumAmount =
       this.minimumAmount ??
--- a/src/calculator/index.ts
+++ b/src/calculator/index.ts
@@ -186,12 +186,22 @@ export default class Calculator {
       round.metadata?.quadraticFundingConfig?.sybilDefense ??
       false;
 
+    // 1. convert threshold amount to 6 tokenDecimals
+    // 2. truncate rest of decimals to bigint
+    // 3. convert decimals to tokendecimals
+    // 4. remove initial 6 decimals
     const minimumAmount =
       this.minimumAmount ??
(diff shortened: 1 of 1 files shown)


src/calculator/index.ts
```diff
@@ -186,12 +186,22 @@ export default class Calculator {
       round.metadata?.quadraticFundingConfig?.sybilDefense ??
       false;
 
+    // 1. convert threshold amount to 6 decimals
+    // 2. truncate rest of decimals to bigint
+    // 3. convert decimals to token decimals
+    // 4. remove initial 6 decimals
     const minimumAmount =
```
hmrtn, line 193: This works as expected - but for readability we could choose to put the nested operations in its own function. Something like: 
```

**Review by hmrtn — APPROVED** [review]

File: src/calculator/index.ts [hunk 2]
```diff
@@ -186,12 +186,22 @@ export default class Calculator {
...
       this.minimumAmount ??
-      BigInt(
-        round.metadata?.quadraticFundingConfig?.minDonationThresholdAmount ?? 0
+      (BigInt(
+        Math.trunc(
+          Number(
```
hmrtn, line 197: This nested bit can be a bit difficult to read and I would not be opposed to breaking this into its own function. Something like: 
```
function calculateMinimumAmount(
  minDonationThresholdAmount: number | undefined,
  matchTokenDecimals: bigint
): bigint {
  if (minDonationThresholdAmount === undefined) {
    return 0n;
  }

  const sixDecimalMultiplier = Math.pow(10, 6);
  const truncatedAmount = Math.trunc(minDonationThresholdAmount * sixDecimalMultiplier);
  const bigIntTruncatedAmount = BigInt(truncatedAmount);

  const decimalConversionFactor = 10n ** matchTokenDecimals / 10n ** 6n;
  const minimumAmount = bigIntTruncatedAmount * decimalConversionFactor;

  return minimumAmount;
}

const minDonationThresholdAmount = round.metadata?.quadraticFundingConfig?.minDonationThresholdAmount;
const minimumAmount = this.minimumAmount ?? calculateMinimumAmount(minDonationThresholdAmount, matchTokenDecimals) ?? 0n;

let matchingCapAmount = this.matchingCapAmount;
```
but thats just my opinion! lgtm :D

vacekj: Merging now so we can test the E2E flow and will refactor Later™

PR #65, opened 2023-05-08
feat: ignore empty override file

vacekj:
only errors out if the override file is actually there, not just when the parameter is set.

### DIFF
--- a/src/test/http/app.test.ts
+++ b/src/test/http/app.test.ts
@@ -229,7 +229,7 @@ describe("server", () => {
         expect(projects.has("project-id-3")).toBe(true);
       });
 
-      test("should render 400 if no overrides file has been uploaded", async () => {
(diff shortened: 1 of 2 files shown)


src/test/http/app.test.ts
```diff
@@ -229,7 +229,7 @@ describe("server", () => {
         expect(projects.has("project-id-3")).toBe(true);
       });
 
-      test("should render 400 if no overrides file has been uploaded", async () => {
+      test.skip("should render 400 if no overrides file has been uploaded", async () => {
```
gravityblast, line 232: we should remove it if we don't want it anymore

boudra (changes requested):
this is not needed anymore

boudra:
closing because it's not needed

PR #66, opened 2023-05-08
Round exports API endpoint

boudra:
- Split up CSV exports into their own group of routes
- Add round votes export
- Add round export
- Add prices export
- Introduces an API `/api/v1/` prefix where we can put all the dynamic routes, we keep the raw data under the `/data` prefix, Solarpunk team are looking into adding some routes so this makes it easy for them to do so
- This also sets the foundation for common error handling in the API, at the moment unexpected errors in all routes will return an internal server error and will be reported to Sentry, you can throw a ClientError inside the routes to return custom errors
- Added backwards compatible routes, they need to be removed after we deploy the frontend

Related PRs: https://github.com/gitcoinco/allo-indexer-client/pull/23 https://github.com/gitcoinco/grants-stack/pull/1752


gravityblast (approved):

src/http/api/v1/exports.ts
```diff
@@ -0,0 +1,333 @@
...
+      "userGithub",
+      ...questionTitles,
+    ],
+  });
+
+  const records = [];
+
+  for (const application of applications) {
+    const answers =
+      application.metadata?.application.answers.flatMap((answer: any) => {
```
hmrtn, line 234: Are we unsure of the shape of answers at this point? 
If not, can we type it: 
```typescript
type Answer = {
  question: string;
  answer: string;
  encryptedAnswer?: boolean;
};
```

src/http/api/v1/exports.ts
```diff
@@ -0,0 +1,333 @@
...
+    ].map((h) => ({
+      id: h,
+      title: h,
+    })),
+  });
+
+  return csv.getHeaderString()!.concat(csv.stringifyRecords([round]));
+}
+
+async function exportApplicationsCSV(db: JsonStorage, round: Round) {
```
hmrtn, line 195: may be annoyingly pedantic and unnecessary, but this function seems like it is doing a lot - extracting question titles, creating a CSV stringifier, and iterating over applications to build the records. Perhaps we could split this up into smaller chunks e.g.
```
function extractQuestionTitles(apps[]): string[]

and 

function buildApplicationRecord(app, titles, etc): records[]
```
That way `exportApplicationsCSV` is only responsible for exporting the CSV.

```typescript
async function exportApplicationsCSV() {
  // ... load from collection ...
  const questionTitles = extractQuestionTitles(apps)
  // ... csv stringifier ...
  const records = apps.map(app => buildApplicationRecord(app, titles, etc)
}
  return csv.getHeaderString() + csv.stringifyRecords(records);

```

if that makes sense at all

src/http/api/v1/exports.ts
```diff
@@ -0,0 +1,333 @@
...
+    fs.promises.readFile("./data/passport_scores.json", {
+      encoding: "utf8",
+      flag: "r",
+    }),
+  ]);
+
+  const passportScores: FullPassportScore[] = JSON.parse(data);
+  const passportScoresMap = processPassportScores(passportScores);
+
+  const records = votes.map((vote: any) => {
```
hmrtn, line 115: Are we still unsure of the shape? `Vote` type is pretty well established at this point, I think. 

```typescript
type Vote = {
  id: string;
  projectId: string;
  applicationId: string;
  roundId: string;
  token: string;
  voter: string;
  grantAddress: string;
  amount: string;
  amountUSD: string;
};
```

src/http/api/v1/matches.ts
```diff
@@ -0,0 +1,120 @@
...
+  // unexpected error, rethrow to upper level handler
+  throw err;
+}
+
+export const calculatorConfig: { dataProvider: DataProvider } = {
+  dataProvider: new FileSystemDataProvider("./data"),
+};
+
+async function matchesHandler(
+  req: Request,
```
hmrtn, line 42: again, totally unnecessary, but possible - we can narrow down the generic `Request` type to be _really_ specific (this goes for _all_ the express Request/Response types. 
```typescript
type MatchRequest = Request<
  { chainId: string; roundId: string },
  any,
  any,
  {
    minimumAmount?: string;
    passportThreshold?: string;
    matchingCapAmount?: string;
    enablePassport?: string;
    ignoreSaturation?: string;
  }
>;
```
such that 
```
async function matchesHandler(
  req: MatchRequest,
  ...
```
etc, etc. again just nitpicking :)

hmrtn (approved):
Overall its good to go! Left comments that I think we can resolve in the future.

hmrtn:
Oh 
its already merged

PR #69, opened 2023-05-09
coefficients export: set default coefficient to 1 if passport is not used

gravityblast:


src/http/api/v1/exports.ts
```diff
@@ -109,14 +109,23 @@ async function exportVoteCoefficientsCSV(db: JsonStorage, round: Round) {
     }),
   ]);
 
+  const usePassport =
```
boudra, line 112: is it a react hook?

boudra (approved):

src/http/api/v1/exports.ts
gravityblast: @boudra  yeah it's defined server side but it runs on a remote browser.

hmrtn: Mo is right! This should be conventionally boolean name like - such as `isPassport` or `isSybilDefended`, etc.

boudra:
whoops build is failing

gravityblast:
@boudra I fixed a eslint error, it's rebuilding

src/http/api/v1/exports.ts
gravityblast: @hmrtn @boudra  this var tells the rest of the program if we need to use passport or not. 
`isPassport` wouldn't make sense.

would you prefer `isPassportEnable`?

boudra: `isPassportEnabled` sounds good

gravityblast: I updated it to `isPassportEnable`

hmrtn (dismissed):

boudra (approved):

PR #77, opened 2023-05-18
Stricter typing and linting

boudra:
- Enable stricter Eslint rules for typing
- Implement all types for passport, calculator and tests

There are no changes to functionality.

depends on https://github.com/gitcoinco/allo-indexer/pull/112


vacekj:
@boudra what's the status on this? I'm down to tag-team on this to get it merged.

boudra:
> @boudra what's the status on this? I'm down to tag-team on this to get it merged.

need to merge this one first: https://github.com/gitcoinco/allo-indexer/pull/112

then this one is ready to go as well

vacekj (approved):

src/config.ts
```diff
@@ -20,7 +19,7 @@ const chains: Chain[] = [
   {
     id: 1,
     name: "mainnet",
-    rpc: `https://mainnet.infura.io/v3/${process.env.INFURA_API_KEY}`,
+    rpc: `https://mainnet.infura.io/v3/${process.env.INFURA_API_KEY ?? ""}`,
```
bard, line 22: Not for this PR but I think that eventually we should have a `getChains` method that returns this object and that throws when the necessary API KEYs aren't provided, so we don't start the service with bad configuration.

bard (approved):

vacekj (approved):

src/config.ts
boudra: @bard that's a great idea, just a note though is that both the Infure and Alchemy API keys are not required, the indexer should still work, might get some more rate limit errors though, but could be worth making them required.

PR #80, opened 2023-05-23
Fix Enable Passport Flag

boudra:
- `sibylDefense` flag in the metadata is not taken into account because `enablePassport` will always be false: https://github.com/gitcoinco/allo-indexer/pull/29
- Passport address index is case sensitive


src/utils/ipfs.ts
```diff
@@ -10,9 +10,12 @@ export async function fetchJsonCached<T>(
   return await cache.lazy<T>(`ipfs-${cid}`, () => fetchJson<T>(cid));
 }
 
+const cidRegex = /^(Qm[1-9A-HJ-NP-Za-km-z]{44}|baf[0-9A-Za-z]{50,})$/;
```
boudra, line 13: added this in here otherwise the build fails because of some malformed ipfs hashes in optimism

PR #82, opened 2023-05-24
Refactor Vote Coefficients

boudra:
Most of the issues we've had with the coefficients has been a design issue, the calculator and the coefficient export have their own implementation, testing the vote coefficients doesn't guarantee the same results in the calculator.

**This PR is a refactor and doesn't change any functionality, I will make another PR to implement floating point coefficient on top of this for easy review.**

Summary of changes:

- Introduce a `getVotesWithCoefficients` source of truth function that will be called from the calculator and the CSV export handler
- Streamline Passport and Calculator types, remove all `any` and Eslint rule exception. Calculator is now fully type safe.
- Refactor vote data augmentation with passport to be more explicit with less spreads
- Refactor coefficient logic to be more explicit and easy to follow
- Rename `minimumAmount` to `minimumAmountUSD` since this naming caused a bit of confusion, this is a functionality change but we're not using it

I am confident this is not breaking anything but I would wait until beta rounds are finalized to merge.

# Topics of discussion for the group review

- I want to put forward this example as a way for us to improve the codebase and implement features at the same time
     - Refactors have to go in a different PR and should not change functionality, ideally done before the feature
     - Smaller refactors could be done within the feature PR if the code is related
     - Use `git add -p` and `git stash` to your advantage
- Let's walk through some of the issues we had in the calculations that this PR fixes.
   - We can see that they were mostly caused by errors in the business logic.
   - We should spend some time thinking about the feature we're implementing, and thinking how we can integrate it in the existing codebase instead of just implementing it on top.
   - Wrongly named variables can cause bugs, let's put more thought into names.

### DIFF
--- a/src/calculator/errors.ts
+++ b/src/calculator/errors.ts
@@ -1,0 +1,29 @@
+export class CalculatorError extends Error {
+  constructor(message?: string) {
+    super(message);
+  }
+}
+
+export class FileNotFoundError extends CalculatorError {
+  constructor(fileDescription: string) {
+    super(`cannot find ${fileDescription} file`);
+  }
+}
+
+export class ResourceNotFoundError extends CalculatorError {
+  constructor(resource: string) {
+    super(`${resource} not found`);
+  }
+}
+
+export class OverridesColumnNotFoundError extends CalculatorError {
+  constructor(column: string) {
+    super(`cannot find column ${column} in the overrides file`);
+  }
+}
+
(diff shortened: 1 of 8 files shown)


gravityblast (approved):

vacekj:
Re-ran failed CI, otherwise LGTM - good work on the refactor.

vacekj (approved):

PR #98, opened 2023-06-08
Fix getBlockByTimestamp for Optimism Bedrock

boudra:
resolves https://github.com/gitcoinco/grants-stack/issues/1925

getBlockByTimestamp is used when we get prices from Coingecko, we build an index of prices to be queried from block number instead of timestamp, to avoid getting timestamps on the voted event.

Up until now we were estimating the block number by calculating the average block speed, which is normally constant, 12 seconds in mainnet for example. Now the issue is that with the Optimism Bedrock update their block speed is now 2 seconds per block, so this calculation doesn't work anymore because the speed is variable.

I reverted back to an old implementation of binary search but with added optimizations, it's actually quite fast!

Summary of changes:

- getBlockByTimestamp receives a margin parameter to indicate how close we want the block timestamp to be to the target timestamp, this way we can tolerate a bit of error but increase the speed considerably, since we're only using this for prices, I made it so we tolerate 30min of error
- getBlockByTimestamp uses binary search to find the block number closes to the timestamp
- getBlockByTimestamp will use the in memory cache to find a closer range for binary search, instead of starting at 0 and latest
- getBlockByTimestamp now uses an explicit cache object instead of a memoized function
- add an initial_data directory which will be used to load historical prices


hmrtn (dismissed):
This is really solid - great work @boudra!

PR #105, opened 2023-06-16
Use storage directory from env var

boudra:
In the Fly VM it's not necessarily in `./data`


gravityblast (approved):
shall we add the STORAGE_DIR var in the .env file so it's clear we can have it?
maybe with a comment to specify its default

bhargavaparoksham (approved):

vacekj (approved):

PR #109, opened 2023-06-21
test: matching cap

### LINKED ISSUE #103: Indexer: Matching Cap Integration Test
- Test the matching cap feature of the matches endpoint
- Take into account that you can enable it through the query parameters and the round settings

You can assert that none of the matches returned pass the matching cap

vacekj:
fixes #103

### DIFF
--- a/src/test/fixtures/rounds.json
+++ b/src/test/fixtures/rounds.json
@@ -14,5 +14,16 @@
         "sybilDefense": true
       }
     }
+  },
+  {
+    "id": "0x3",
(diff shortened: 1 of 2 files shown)


boudra (commented):
we need to test that matchingCap is disabled as well, or is that covered by other tests?

vacekj:
> 

I think that should be covered by the "pure" matching test?

vacekj:
@boudra added a test for disabled matching cap.

boudra (approved):
🚀

bhargavaparoksham (approved):

PR #111, opened 2023-06-22
Update dev script name

boudra:
I initially named the dev script `start:dev` with the idea that all scripts are prod by default and then you add the `:dev` suffix to make it development.

But the common naming convention for the main dev script is just `npm run dev`, so I propose we use that to make it more obvious to everyone and be consistent with the node ecosystem.


bard (approved):

bard:
@boudra in case it comes up — `foo:watch` might be a good name for the remaining scripts (e.g. `test:dev` → `test:watch` etc)

PR #124, opened 2023-06-27
Refactor for testability

bard:
Initial / exploratory implementation of the proposal at https://github.com/gitcoinco/allo-indexer/issues/119


src/cli/indexer-cli-args.ts
```diff
@@ -0,0 +1,102 @@
...
+      default:
+        throw new Error("Invalid log level.");
+    }
+  }
+
+  const storageDir = path.join(
+    config.storageDir,
+    // XXX could probably switch to static chainName -> chainId mappings so as not to rely on provider.getNetwork() and
+    // not have to make this function async
+    `${provider.network.chainId}`
```
boudra, line 82: I believe we should be able to do this:

```suggestion
    `${chain.id}`
```

bard:
Superseded.

PR #125, opened 2023-06-28
Improve Start Script Readibility

boudra:
I hope I am forgiven for one last command refactor:

- Split out the horrendous start command into a bash script so it can be more readable and easier to understand what the application is doing in production (yes we're going back to a bash script 🙃 )
- Add `--restart-tries` option to Concurrently so that it restarts scripts when they die, I thought this was a default option. I thought 10 was a sensible option.
- Remove the `index:all` command and do it manually instead, more code but it should hopefully be easier to read
- Rename `:dev` commands to watch which makes what the commands do clearer

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+set -e
+
+# Exit script on interrupts
+trap "exit 130" INT
+
+echo "======> Catching up"
+
+# Catch up indexers to latest block before starting
+npx concurrently \
+  --kill-others-on-fail \
+  --names "mainnet,optimism,goerli,fantom,passport" \
+  'npm:index -- --chain=mainnet --log-level=error --clear' \
+  'npm:index -- --chain=optimism --log-level=error --clear' \
+  'npm:index -- --chain=goerli --log-level=error --clear' \
+  'npm:index -- --chain=fantom --log-level=error --clear' \
+  'npm:passport'
+
+echo "======> Catch up successful, running indexer on follow mode!"
+
+# Once caught up, start indexers in follow mode and run HTTP server
+exec npx concurrently \
+  --restart-retires=10 \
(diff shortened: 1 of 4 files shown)


vacekj:
Nothing to forgive - I like that you take care to polish even "small" things like npm scripts, because it shows you care a lot about the product 🤝

vacekj (approved):

start.sh
```diff
@@ -0,0 +1,34 @@
...
+# Once caught up, start indexers in follow mode and run HTTP server,
+# we can now start serving traffic
+exec npx concurrently \
+  --restart-tries=10 \
+  --names "mainnet,optimism,goerli,fantom,passport,http" \
+  'npm:index -- --chain=mainnet --follow' \
+  'npm:index -- --chain=optimism --follow' \
+  'npm:index -- --chain=goerli --follow' \
+  'npm:index -- --chain=fantom --follow' \
+  'npm:passport -- --restart--retries=10 --follow' \
```
bard, line 33: Is this passing `--restart--retries=10` to `passport`?

boudra: good catch!

start.sh
```diff
@@ -0,0 +1,34 @@
...
+  'npm:index -- --chain=goerli --log-level=error --clear' \
+  'npm:index -- --chain=fantom --log-level=error --clear' \
+  'npm:passport'
+
+echo "======> Catch up successful, running indexer on follow mode!"
+
+# Once caught up, start indexers in follow mode and run HTTP server,
+# we can now start serving traffic
+exec npx concurrently \
+  --restart-tries=10 \
```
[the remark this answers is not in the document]
boudra: For catch up it makes sense to kill  all the other processes if one fails because we cannot continue, but when indexing live events we just want to restart a process if it dies, without killing the others. Or would you kill the other processes if one fails?

bard (approved):

PR #128, opened 2023-06-29
Deterministic Event Handler

boudra:
This deletes thunks so that events are always processed sequentially, this should always yield the same result with consequent runs.

```bash
$ npm run index -- --chain=mainnet --clear
$ mv data/1 snapshot_1
$ npm run index -- --chain=mainnet --clear
$ diff --brief -r snapshot_1 data/1
Files snapshot_1/_subscriptions.json and data/1/_subscriptions.json differ
Files snapshot_1/prices.json and data/1/prices.json differ
```

Subscriptions and prices are different, that is expected.

There's a small behaviour change from main (happy to put this in a separate PR):

- If the metadata pointer is invalid, we still update the `metaPtr` but we reset the metadata to `null`, instead of keeping the old metadata, which could be confusing.

### DIFF
--- a/src/indexer/handleEvent.ts
+++ b/src/indexer/handleEvent.ts
@@ -82,36 +82,19 @@ async function handleEvent(
         event.address
       );
 
-      try {
-        await db.collection<Project>("projects").updateOneWhere(
-          (p) => p.projectNumber === event.args.projectID.toNumber(),
-          (project) => ({
-            ...project,
-            metaPtr: event.args.metaPtr.pointer,
-          })
-        );
+      const metadata = await ipfs<Project["metadata"]>(
+        event.args.metaPtr.pointer,
+        indexer.cache
+      );
 
-        return async () => {
-          const metadata = await ipfs<Project["metadata"]>(
-            event.args.metaPtr.pointer,
-            indexer.cache
-          );
-
-          if (!metadata) {
-            return;
-          }
-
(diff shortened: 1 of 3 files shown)


src/indexer/handleEvent.ts
```diff
@@ -436,102 +419,126 @@ async function handleEvent(
 
     // --- Votes
     case "Voted": {
-      return async () => {
-        const voteId = ethers.utils.solidityKeccak256(
-          ["string"],
-          [`${event.blockNumber}-${event.logIndex}`]
-        );
+      const voteId = ethers.utils.solidityKeccak256(
```
boudra, line 422: this event handler diff got all messed up, all  we're doing is removing the surrounding async closure, the rest is formatting.

src/indexer/handleEvent.ts
```diff
@@ -550,97 +557,70 @@ async function handleEvent(
...
 
-        await db
-          .collection<Application>(
-            `rounds/${event.args.roundAddress}/applications`
+      const contributorPartitionedPath = vote.voter
+        .split(/(.{6})/)
+        .filter((s: string) => s.length > 0)
+        .join("/");
+
+      await Promise.all([
```
bard, line 583: Since `Promise.all` runs promises in parallel, given the nature of these operations, is there a chance that a different order of resolution might result in different db state? That shouldn't be the case with `updateById`, I'm mostly wondering about `insert`.

PR #142, opened 2023-07-09
matching estimates

vacekj:
frontend pr: https://github.com/gitcoinco/grants-stack/pull/2022
ticket: https://github.com/gitcoinco/allo-indexer/issues/307


src/calculator/index.ts
```diff
@@ -266,10 +277,138 @@ export default class Calculator {
...
+      "votes",
+      `${this.chainId}/rounds/${this.roundId}/votes.json`
+    );
+
+    const rounds = await this.parseJSONFile<Round>(
+      "rounds",
+      `${this.chainId}/rounds.json`
+    );
+
+    const round = rounds.find((r: Round) => r.id === this.roundId);
```
bard, line 299: The type hint _looks_ like it could be omitted.

src/calculator/index.ts
```diff
@@ -266,10 +277,138 @@ export default class Calculator {
...
+
+    const applications = await this.parseJSONFile<Application>(
+      "applications",
+      `${this.chainId}/rounds/${this.roundId}/applications.json`
+    );
+
+    const currentResults = await this._calculate(votes);
+    const potentialVotesAugmented: Vote[] = await Promise.all(
+      potentialVotes.map(async (vote) => {
+        const { amount: amountUSD } = await this.priceProvider.convertToUSD(
```
bard, line 321: Just something to think about, but a`getUsdConversionRate(chainId, token)` in the price provider would allow to make one async call only outside of this loop instead of `potentialVotes.length * 3` async calls. /cc @boudra

src/http/api/v1/matches.ts
```diff
@@ -111,5 +113,61 @@ export const createHandler = (config: HttpApiConfig): express.Router => {
...
+
+  async function estimateMatchesHandler(
+    req: Request,
+    res: Response,
+    okStatusCode: number
+  ) {
+    const chainId = Number(req.params.chainId);
+    const roundId = req.params.roundId;
+    const potentialVotes = potentialVotesSchema
+      .parse((req.body as { potentialVotes: PotentialVotes }).potentialVotes)
```
bard, line 128: The cast here shouldn't be necessary as the type is taken care of by zod.

src/http/api/v1/matches.ts
```diff
@@ -111,5 +113,61 @@ export const createHandler = (config: HttpApiConfig): express.Router => {
...
+      .map((vote) => ({
+        ...vote,
+        amount: BigInt(vote.amount),
```
bard, line 131: Is this still necessary after using `coerce.bigint()` in the zod schema?

src/test/http/app.test.ts
```diff
@@ -322,6 +325,97 @@ describe("server", () => {
         expect(resp.statusCode).toBe(200);
         expect(resp.body).toEqual(expectedResults);
       });
+
+      test.only("should estimate matching with new votes for projects", async () => {
```
bard, line 329: The `only` here looks like a leftover. (Another good candidate for the linter, incidentally.)

bard (dismissed):

src/calculator/index.ts
boudra: Yes, we should fix this, we had a similar problem in the past where we were looping through all the Passports for each vote. There's already a `getUSDConversionRate`, it looks like the process of then converting the `bigint` amount to USD (floating point) is not straightforward, but that can be extracted into it's own function.

src/http/api/v1/matches.ts
```diff
@@ -111,5 +113,61 @@ export const createHandler = (config: HttpApiConfig): express.Router => {
...
+      }));
+
+    const chainConfig = config.chains.find((c) => c.id === chainId);
+    if (chainConfig === undefined) {
+      throw new Error(`Chain ${chainId} not configured`);
```
boudra, line 136: This will return a 500 error, it should return a 4xx error to the client, `ClientError` can be used for that.

src/calculator/index.ts
```diff
@@ -266,10 +277,138 @@ export default class Calculator {
...
+            this.chainId,
+            round.token,
+            amountUSD
+          );
+
+        /* Find the latest approved application */
+        const application = applications
+          .filter(
+            (application) =>
+              application.metadata?.application.recipient === vote.recipient
```
gravityblast, line 338: we can't find the application by recipient, since many applications/projects might have the same recipient. 

Instead we should pass from the front-end what we send to the voting strategy, which is exactly what we already index and have in the current votes:

```
        address _token,
        uint256 _amount,
        address _grantAddress,
        bytes32 _projectId,
        uint256 _applicationIndex
```

If we do this we don't need to load and validate rounds and applications here above, since it's are already done in `_calculate`.

src/calculator/index.ts
```diff
@@ -266,10 +277,138 @@ export default class Calculator {
...
+          )
+          .filter((application) => application.status === "APPROVED")
```
gravityblast, line 340: We can skip the check on the application status because we are not indexing but just simulating the results on demand for the current request.
If a client sends an application not approved it will get back wrong results, bad for itself but it's not saved anywhere, and since we are building the request in Grants Stack, we know the response will be good for us.

src/calculator/index.ts
```diff
@@ -266,10 +277,138 @@ export default class Calculator {
...
+    }
+
+    if (round.token === undefined) {
+      throw new ResourceNotFoundError("round token");
+    }
+
+    const applications = await this.parseJSONFile<Application>(
+      "applications",
+      `${this.chainId}/rounds/${this.roundId}/applications.json`
+    );
```
gravityblast, line 316: we can skip all these validations since they are already done in `_calculate`. 
It's better to have a single point with this logic, if we add more validations in `_calculate` we might forget to add them here as well.

src/calculator/index.ts
gravityblast: I agree, we can add it directly there or having it here getting a price with a constant amount to calculate the rate.

gravityblast (changes requested):

src/calculator/index.ts
vacekj: Good point - fixed and renamed the r to round to be more clear.

src/http/api/v1/matches.ts
vacekj: True - missed this. fixed.

src/http/api/v1/matches.ts
vacekj: Nope - good catch. I see that I'm rusty on zod heh.

src/test/http/app.test.ts
vacekj: Thanks - fixed and added an eslint plugin and rule to forbid `only`s

src/http/api/v1/matches.ts
vacekj: Yep, thank you - fixed to throw a clientError with status 400.

src/calculator/index.ts
```diff
@@ -139,11 +141,39 @@ export default class Calculator {
...
   async calculate(): Promise<Array<AugmentedResult>> {
     const votes = await this.parseJSONFile<Vote>(
       "votes",
       `${this.chainId}/rounds/${this.roundId}/votes.json`
     );
+
+    return this._calculate(votes);
+  }
+
+  private async _calculate(votes: Vote[]): Promise<Array<AugmentedResult>> {
```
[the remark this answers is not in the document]
vacekj: `private` is a typescript feature, `#` is an ECMAScript spec thing. I'm generally for upholding the spec instead of relying on typescript where possible. Good discussion here: https://stackoverflow.com/questions/59641564/what-are-the-differences-between-the-private-keyword-and-private-fields-in-types.

src/calculator/index.ts
vacekj: Very good point - will fix.

src/calculator/index.ts
vacekj: Good idea - fixed.

src/calculator/index.ts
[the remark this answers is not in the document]
vacekj: temp. fixed to private so as not to block this PR over style discussions.

src/calculator/index.ts
vacekj: Fixed.

vacekj:
@boudra @bard there are two test failures that I'm not immediately sure how to resolve, I'll investigate tomorrow but it might be a quick fix for you.

bard:
@vacekj I've narrowed it down to `tmpNameSync` returning the same "temporary" (not so temporary) name twice in a row. Can be seen by `console.log`ging inside `PassportProvider.start`. Switching to `tmpName` (async version) didn't help.

Cannot look into it further right now but to debug, you can `console.log` the `dbPath` inside `PassportProvider.start` — as long as it prints the same name more than once, the problem is there. To fix, `tmpNameSync` in `src/passport/index.test.fixtures.ts` needs to be replaced with something that actually generates different names.

boudra:
I can look into it in half an hour, but if it helps i had this exact same problem in my branch when i updated Vitest, downgrading it fixed the problem, very strange.

boudra:
I pushed a commit to rollback Vitest to 0.34.5 so that we can leave things as they were and we can make fix the passport issue separately since it might not be an obvious fix.

`tmpNameSync` seems to definitely be generating random paths, the issue I think is with `test.extend` which seems to be giving each test the same context instead of generating one every time, which maybe is a behaviour change or bug that was introduced in Vitest 0.34.6.

boudra (approved):

gravityblast (approved):

Issue #155, opened 2023-07-14
service unresponsive during deployment


bard:
The service is slow to start since it currently re-indexes from scratch on startup, but still this should not happen — we have rolling upgrades in fly.io.

Issue #154, opened 2023-07-13
add logging


bard:
We need multi-level logging for debugging as well as to help newcomers to the codebase to trace the execution. [pino](https://www.npmjs.com/package/pino) (plus [pino-pretty](https://www.npmjs.com/package/pino-pretty) for local use) are popular and well-tested choices.

PR #157, opened 2023-07-16
refactor/fix: merge services, add logging, fix goerli, fix price edge case, unify config, rewrite passport updater

bard:
Sorry for the kitchen-sink PR but, as someone said, you can't cross a chasm in two small jumps.

Summary:

  - [func,fix] fix price aliasing for goerli (BUSD in existing events)
  - [func,fix] handle edge case in price correlation (requested block is older than earliest available block number for given token)
  - [ops,enh] reduce start time from >25m to <5m by using existing passport dataset if available
  - [ops,enh] add machine-readable structured logging
  - [ops,enh] patch chainsauce to use external logger
  - [ops,feat] ship logs to DataDog
  - [func,enh] replace ad-hoc `fetchRetry` with [make-fetch-happen](https://www.npmjs.com/package/make-fetch-happen) library that provides retrying, caching, and more
  - [func,enh] replace ad-hoc atomic file write routine with [write-file-atomic](https://www.npmjs.com/package/write-file-atomic) library
  - [dev,enh] remove remaining module-level side effects (config parsing, service instantiations)
  - [dev,feat] rewrite passport updater from scratch in preparation for incremental updates (depends on passport API changes)
  - [dev,feat] move all functionality (passport updater, per-chain price+data updater, http api) into single process; remove drop script; unify config
  - [dev,enh] make remaining dependencies injectable (cache, contract subscription, data provider for http api); decouple `handleEvent` from chainsauce indexer

Structured logs:

image

Screenshot from logging dashboard:

image

Closes: #155, #154, #135,


src/prices/provider.ts
```diff
@@ -0,0 +1,184 @@
+import { Logger } from "pino";
+import { CHAINS } from "../config.js";
+import { Price, readPricesFile } from "./common.js";
+
+interface PriceProviderConfig {
+  updateEvery?: number; // XXX hint at time type: secs? msecs?
```
vacekj, line 6: this would be great - i've seen a `ms` suffix used in the wild, so this would become `updateEveryMs`.

bhargavaparoksham (approved):

PR #161, opened 2023-07-18
poc: simple unit test example

bard:
Not for merging.

This shows a (working) unit test for the simplest event handler I could find. It relies on all the dependency extraction done in the past weeks.

I don't think the approach of mocking "just enough" of the db object is scalable to more complex event handlers. 

An alternative would be to patch chainsauce so that it accepts a `node:fs` implementation, and use [memfs](https://www.npmjs.com/package/memfs) during tests. 

Or store to SQL, as Mo already wanted to do, which can then be tested using sqlite in-memory database, and which would involve writing an adapter: https://github.com/chainsauce-org/chainsauce/blob/6fcec0e98eb508953a829dab1e6fca636cc728de/src/storage/json.ts

Opinions?

### DIFF
--- a/src/indexer/handlers/roundMetaPtrUpdated.test.ts
+++ b/src/indexer/handlers/roundMetaPtrUpdated.test.ts
@@ -1,0 +1,62 @@
+import { describe, test, expect, vitest } from "vitest";
+import { default as roundMetaPtrUpdated } from "./roundMetaPtrUpdated.js";
+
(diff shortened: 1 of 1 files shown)


vacekj:
@gitcoinco/grantsstack-engineering are there any actions left here, or are we good to close?

bard:
It was informational only and I think @boudra looked at this already, thus good to close.

PR #177, opened 2023-07-21
fix: don't hardcode deployment environment in logs

bard:
Required for deployment to production.

fixes: #176


vacekj (approved):
+1 for zod (or any good validation library) 🤝

gravityblast (approved):

bard:
> +1 for zod (or any good validation library) handshake

To be fair, there should be a single schema instead of scattered validations, but given the risk for breakage I went for the incremental change, the idiomatic version is for the next round.

PR #210, opened 2023-08-07
enh: run prettier and lint in git hooks

### LINKED ISSUE #209: add pre-commit and pre-push hooks
pre-commit: prettier
pre-push: lint, test

Use [lefthook](https://github.com/evilmartians/lefthook) for consistency with grants stack.

bard:
closes #209 

Will decrease diff noise by ensuring that all contributions are formatted to the same style, and history noise by preventing commits with lint issues to be pushed.

The `trailingComma` setting in `.prettierrc.json` is due to the default having changed in 3.0. We can switch to the new default in a separate PR over the entire codebase, again to prevent noise.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,10 +14,6 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - name: Write commit hash to .env
-        run: |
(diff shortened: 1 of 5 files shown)


vacekj (approved):
lgtm, agree we can run the reformat with no config (and therefore trailing commas everywhere) in another pr. personally it's more comfortable for obvious reasons, but the more important thing is keeping the config the default.

PR #213, opened 2023-08-07
enh: import abis statically

bard:
Move reading of ABIs to compile time to catch problems sooner and reduce async code.

Based off https://github.com/gitcoinco/allo-indexer/pull/208 which should be merged first.

Supersedes #174 and #171 except for AIP-13 support, which should be added separately [TODO add example of synthetic ABI]


vacekj (approved):

bard:
closed and reworked in #275 due to extensive conflicts

PR #259, opened 2023-08-24
apply vote amount cap to votes

gravityblast:
closes https://github.com/gitcoinco/allo-indexer/issues/260

Each vote is passed to `applyVoteCap`; the amount is capped if in the settings we have a cap for a vote chainId/tokenAddress.

We apply the cap to `amountRoundToken`, which is the amount passed to the `linearQF` function.


bard (approved):

michellema1208:
this can be merged to production whenever ready!

PR #290, opened 2023-09-19
enh: initial catchup performance

boudra:
This aims to fix the current blocking issue with deployments by reducing the time it takes to do an initial catch up without making too many changes.

**Note: There's a temporary change to auto deploy this branch to staging to test deployments, remove before merging.**

Depends on https://github.com/gitcoinco/chainsauce/pull/2 https://github.com/gitcoinco/chainsauce/pull/1

## IPFS

Some IPFS files do not seem to exist anymore, I can see them on the Pinata dashboard but the IPFS server never returns a response, this happens even when using different gateways, example: https://ipfs-grants-stack.gitcoin.co/ipfs/bafkreig47gxcq5j2mkligw2odnitqqhxfhs34mz5r6yry4jtfgqr54f6yq

This holds the deployments due to high timeouts in the IPFS fetch and finally crashes the indexer when it fails. We could treat it as a warning as the hash comes from the user.

## RPC (Chainauce)

Chainsauce's RPC fetching depends on `node-fetch` and this issue was affecting us: https://github.com/node-fetch/node-fetch/issues/1735

The retry logic is also a bit aggressive and it could overwhelm smaller servers like Fantom or PGN testnet.

## Storage access by ID (Chainsauce)

On each vote we check whether the contributor has already donated or not, we do this by looking up the contributor address, this can get expensive as the search is linear and the number of contributors can be very high, this also happens inside the tightest loop of the indexing process so it affects the indexing time considerably, Chainsauce now implements and index to cover cases when we want to find a record by ID.

## Memory leak (Chainsauce)

Fixed a bug where Chainsauce kept all the JSON files in memory, they're now be garbage collected as intended when written to the disk. The indexer should use a lot less memory now, looks like production is at 3GB and staging at 1GB running the same chains.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
   "license": "ISC",
   "dependencies": {
     "@sentry/node": "^7.51.0",
-    "chainsauce": "github:gitcoinco/chainsauce#49bee375cc230097ef024c0f7107c13393f5c2ac",
+    "chainsauce": "github:gitcoinco/chainsauce#main",
     "cors": "^2.8.5",
     "csv-parser": "^3.0.0",
     "csv-writer": "^1.6.0",
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
   "license": "ISC",
   "dependencies": {
     "@sentry/node": "^7.51.0",
-    "chainsauce": "github:gitcoinco/chainsauce",
+    "chainsauce": "github:gitcoinco/chainsauce#49bee375cc230097ef024c0f7107c13393f5c2ac",
     "cors": "^2.8.5",
     "csv-parser": "^3.0.0",
     "csv-writer": "^1.6.0",
--- a/src/index.ts
+++ b/src/index.ts
@@ -190,18 +190,32 @@ async function catchupAndWatchChain(
         return undefined;
       }
 
-      const res = await fetch(`${config.ipfsGateway}/ipfs/${cid}`, {
-        timeout: 2000,
-        retry: { retries: 10, maxTimeout: 60 * 1000 },
-        // IPFS data is immutable, we can rely entirely on the cache when present
-        cache: "force-cache",
-        cachePath:
-          config.cacheDir !== null
-            ? path.join(config.cacheDir, "ipfs")
-            : undefined,
-      });
+      const url = `${config.ipfsGateway}/ipfs/${cid}`;
 
-      return (await res.json()) as T;
+      chainLogger.trace(`Fetching ${url}`);
+
+      try {
+        const res = await fetch(url, {
+          timeout: 2000,
+          onRetry(cause) {
+            chainLogger.debug(`Retrying IPFS request ${String(cause)}`);
+          },
+          retry: { retries: 3, minTimeout: 2000, maxTimeout: 60 * 10000 },
+          // IPFS data is immutable, we can rely entirely on the cache when present
+          cache: "force-cache",
+          cachePath:
+            config.cacheDir !== null
+              ? path.join(config.cacheDir, "ipfs")
+              : undefined,
+        });
+
+        return (await res.json()) as T;
+      } catch (err) {
+        chainLogger.warn({
+          msg: "failed to load IPFS file",
+          err,
+        });
+      }
     };
 
     await rpcProvider.getNetwork();
--- a/src/index.ts
+++ b/src/index.ts
@@ -194,28 +194,25 @@ async function catchupAndWatchChain(
 
       chainLogger.trace(`Fetching ${url}`);
 
-      try {
-        const res = await fetch(url, {
-          timeout: 2000,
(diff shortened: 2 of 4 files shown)


.github/workflows/deploy-branch.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         run: |
           echo "DEPLOYMENT_ENVIRONMENT=staging" >> $GITHUB_ENV
           echo "DEPLOYMENT_URL=https://indexer-staging.fly.dev" >> $GITHUB_ENV
-        if: ${{ github.ref == 'refs/heads/main' }}
+        if: ${{ github.ref == 'refs/heads/fix-deployments' }}
```
bard, line 21: Can we test deployments from local instead of CI?

src/index.ts
```diff
@@ -188,18 +189,32 @@ async function catchupAndWatchChain(
...
-      return (await res.json()) as T;
+      const url = `${config.ipfsGateway}/ipfs/${cid}`;
+
+      chainLogger.trace(`Fetching ${url}`);
+
+      try {
+        const res = await fetch(url, {
+          timeout: 2000,
+          onRetry(cause) {
+            chainLogger.debug(`Retrying IPFS request ${String(cause)}`);
```
bard, line 200: You could preserve more information about the error with something like:

```ts
chainLogger.debug({ msg: `Retrying IPFS request`, err: cause })
```

That should make the stack trace available in the log.

bard:
Snapshot testing against `main` is passing.

src/index.ts
```diff
@@ -188,18 +189,32 @@ async function catchupAndWatchChain(
...
+          cachePath:
+            config.cacheDir !== null
+              ? path.join(config.cacheDir, "ipfs")
+              : undefined,
+        });
+
+        return (await res.json()) as T;
+      } catch (err) {
+        chainLogger.warn({
+          msg: "failed to load IPFS file",
```
bard, line 214: Is the `url` shown in the error already? If not, it would be good to have it in the log message.

bard (changes requested):

bard:
@boudra this https://github.com/node-fetch/node-fetch/issues/1735 says it's only Node 19+ but we're using Node 18. Is the issue incorrect? It would be good to know for sure that those ECONNRESET are addressed

src/index.ts
boudra: @bard I didn't realize we caught errors from the handle event, so this catch and log might not be needed here.

This is how the Fetch error I've been seeing looks like serialized so there's enough information to see what's happening:

```
{
      "type": "FetchError",
      "message": "network timeout at: https://ipfs-grants-stack.gitcoin.co/ipfs/bafkreig7gzi3mlj5bcfbrzakisvchdnpok7blwvhs54dsjfcavbct6yrwi",
      "stack":
          FetchError: network timeout at: https://ipfs-grants-stack.gitcoin.co/ipfs/bafkreig7gzi3mlj5bcfbrzakisvchdnpok7blwvhs54dsjfcavbct6yrwi
              at Timeout.<anonymous> (/Users/moboudra/dev/gitcoin/allo-indexer/node_modules/minipass-fetch/lib/index.js:108:18)
              at listOnTimeout (node:internal/timers:564:17)
              at process.processTimers (node:internal/timers:507:7)
      "code": "FETCH_ERROR",
      "errno": "FETCH_ERROR"
}
```

src/index.ts
boudra: done, normally the URL is on the error but added the URL as well in the log just in case

boudra:
> @boudra this [node-fetch/node-fetch#1735](https://github.com/node-fetch/node-fetch/issues/1735) says it's only Node 19+ but we're using Node 18. Is the issue incorrect? It would be good to know for sure that those ECONNRESET are addressed

I am definitely not seeing as frequently them anymore, I've noticed some from Fantom but the new retry logic just retries that single request instead of splitting a range into two requests, making the problem worse: https://github.com/gitcoinco/chainsauce/pull/1

src/index.ts
bard: Yeah, unfortunately "just crash" was untenable in production when the cause of the crash was guaranteed to appear on the next start.

package.json
```diff
@@ -29,7 +29,7 @@
   "license": "ISC",
   "dependencies": {
     "@sentry/node": "^7.51.0",
-    "chainsauce": "github:gitcoinco/chainsauce",
+    "chainsauce": "github:gitcoinco/chainsauce#49bee375cc230097ef024c0f7107c13393f5c2ac",
```
bard, line 32: This is not really needed as the hash is already saved in the lockfile, i.e. there is no danger of installing a version other than the correct one for a given commit.

bard (changes requested):

PR #302, opened 2023-09-27
Fix getBlockFromTimestamp

### LINKED ISSUE #252: investigate block estimation occasionally reporting negative values
https://gitcoin-protocol.sentry.io/issues/4401322787/?project=4505130976870400&query=is%3Aunresolved&referrer=issue-stream&stream_index=1

### LINKED ISSUE #278: intermittent error with block estimation
Impact: blocking deployment.

boudra:
- `getBlockFromTimestamp` is now correct and deterministic
- Uses a local SQLite cache that is easier to manage, more portable and will consume less disk space & inodes
- The initial prices index will now take much longer than before since we're finding the block with a binary search instead of estimating, but once the cache is warm the re-index will take seconds, my local SQLite cache can be used to seed Fly instances or dev machines

Summary of changes:

- Revert the `getBlockFromTimestamp` to a simple unoptimized binary search, remove estimations
- Make `getBlockFromTimestamp` deterministic, it now always returns the _lowest_ block that is equal to the timestamp or is immediately after it, before we were just getting the closest block, but depending on how the binary search went you could get different results with the same inputs, also the Optimism chain a lot of the blocks have the same timestamp which complicates things
- Introduce a "block cache" that stores information about blocks and their timestamps, to be used on reindexes
- There's a first implementation of a block cache using plain`better-sqlite3`

closes #252
closes #278


.eslintrc.cjs
```diff
@@ -12,6 +12,7 @@ module.exports = {
   root: true,
   rules: {
     "no-unused-vars": "off",
+    "@typescript-eslint/require-await": "off",
```
boudra, line 15: Not sure how helpful this rule is, it becomes impossible to implement an interface that requires an async method but the implementation is sync.

bard: What happens if you precede the async call with `void`?

bard: And/or provide a `.then` and a `.catch`.

boudra: I'll try the void, what I've done before in Chainsauce wrap the return value a promise with `Promise.resolve` but it feels like I am adding unnecessary code just to make the linter happy

src/blockCache/sqlite.test.ts
```diff
@@ -0,0 +1,86 @@
+import { createSqliteBlockCache } from "./sqlite.js";
```
bard, line 1: One thing probably worth keeping in mind when using a SQL database is that, because of the consistency requirements, it might enforce sequential writing even for data with no dependencies, and thus be a bottleneck to a degree, compared to a non-relational store.

That said, doing queries with ordering and comparisons would be trickier in non-relational stores.

src/blockCache/sqlite.ts
```diff
@@ -0,0 +1,124 @@
...
+    async getBlockByNumber(chainId, blockNumber) {
+      if (!db) {
+        throw new Error("SQLite database not initialized");
+      }
+
+      const row = db
+        .prepare(
+          `SELECT * FROM ${tableName} WHERE chainId = ? AND blockNumber = ?`
+        )
+        .get(chainId, blockNumber.toString()) as Row | undefined;
```
bard, line 68: Casting might be minimized here and elsewhere and flow made explicit with something like

```ts
if (row === undefined) {
  return null
} else {
  return ... as Block
}

File: src/blockCache/sqlite.ts [hunk 4]
```diff
@@ -0,0 +1,124 @@
...
+        : null;
+    },
+
+    async getBlockByTimestamp(chainId, timestamp) {
+      if (!db) {
+        throw new Error("SQLite database not initialized");
+      }
+
+      const row = db
+        .prepare(
```
bard, line 80: It is possible to prepare a statement once, upon initializing, and use the prepared statement in the call.

src/prices/updater.ts
```diff
@@ -146,26 +156,42 @@ export function createPriceUpdater(
...
-                lastBlockNumber,
-                getBlockTimestamp,
-                logger
-              );
+              const timestampInSeconds = Math.floor(timestampMs / 1000);
+
+              let blockNumber =
+                timestampToBlockMap.get(timestampInSeconds) ?? null;
+
+              if (!blockNumber) {
```
bard, line 166: This will evaluate to `true` both when `blockNumber`is missing from the map, and in case `blockNumber` is 0 (assuming that can ever happen).

src/prices/updater.ts
```diff
@@ -107,15 +118,14 @@ export function createPriceUpdater(
         return;
       }
 
-      const getBlockTimestamp = async (blockNumber: number) => {
-        const block = await withCacheMaybe(
-          `block-${config.chain.id}-${blockNumber}`,
-          () => provider.getBlock(blockNumber)
-        );
+      const getBlockTimestamp = async (blockNumber: bigint) => {
+        const block = await provider.getBlock(Number(blockNumber));
```
bard, line 122: Querying the cache here and falling back to the provider if block is not cached would avoid pushing the caching concern down to `getBlockFromTimestamp`.

src/utils/getBlockFromTimestamp.ts
```diff
@@ -1,165 +1,76 @@
...
-  logger: Logger
-): Promise<number> {
-  const blockEstimationThreshold = 10000;
-  const targetTimestamp = Math.floor(timestampInMs / 1000);
+      const timestamp = await getBlockTimestamp(blockNumber);
 
-  // by default we binary search from 0 to the current block number
-  let start = startBlock;
-  let end = endBlock;
+      // console.log("saving");
```
bard, line 28: Leftover?

src/utils/getBlockFromTimestamp.ts
```diff
@@ -1,165 +1,76 @@
...
-
-  return estimatedBlockNumber;
-};
+  chainId: number;
+  timestampInSeconds: number;
+  startBlock: bigint;
+  endBlock: bigint;
+  getBlockTimestamp: (blockNumber: bigint) => Promise<number>;
+  blockCache?: BlockCache;
+}): Promise<bigint | null> {
```
bard, line 17: Does any sanity check need to be done here to warn about invalid arguments sneaking in or is done elsewhere?

bard (commented):
Approving by default, decision about what to pick and what to leave up to the author.

bard (dismissed):

package.json
```diff
@@ -29,6 +29,8 @@
   "license": "ISC",
   "dependencies": {
     "@sentry/node": "^7.51.0",
+    "@types/better-sqlite3": "^7.6.5",
```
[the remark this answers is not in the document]
boudra: fixed

.eslintrc.cjs
boudra: > To go back to the original question: removing this rule would mean risk of forgetting an `await`, and forgetting an `await` would mean risk of `unhandledRejection`.

You might be thinking of this one https://typescript-eslint.io/rules/no-floating-promises/

This rule prevents you from having a async functions that have no await:

```
// error!
async function test() {
   return 50;
}
```

Which is fine, because this function should not be async, it can be modified, the problem i encountered is having an interface that requires an async method, but the implementation is completely sync, maybe best to add an explicit disable for this scenario since it's not very common.

src/prices/updater.ts
boudra: good catch

.eslintrc.cjs
bard: You're right, I was actually thinking of that rule.

I've encountered this before when I knew an implementation would eventually need to be async, but the proof-of-concept was sync, and I also solved it with a `Promise.resolve()`. Personally I think that's fine.

src/blockCache/sqlite.ts
```diff
@@ -0,0 +1,124 @@
...
+interface Row {
+  chainId: number;
+  blockNumber: string;
+  timestamp: number;
+}
+
+export function createSqliteBlockCache(opts: Options): BlockCache {
+  let db: Sqlite.Database | null = null;
+
+  if (opts.tableName != null && opts.tableName.match(/[^a-zA-Z0-9_]/)) {
```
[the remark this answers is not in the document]
boudra: fixed

bard (approved):

vacekj (approved):

PR #303, opened 2023-09-27
feat: add avalanche support

bard:
For local development, set `AVALANCHE_RPC_URL` if necessary.

Development checklist:

  - [x] verify whether in addition to AVAX we need to support USDC ([PR says so](https://github.com/gitcoinco/grants-stack/pull/2351/files#diff-c61c1dd84668161c5405dbae406b8e48841a6317abc360cecac2448e5af5d01cR396)) or DAI+wETH or DAI+ETH ([ticket says so](https://github.com/gitcoinco/grants-stack/issues/2308)
  - [x] verify whether AVAX price retrieval is implemented correctly (see TODO's in chain support checklist for grey areas)
  - [x] update chain support checklist with instructions for token setup, in particular with regard to CoinGecko configuration in `src/prices/coinGecko.ts`
  - [x] repeat for Fuji

Deployment checklist:
  
  - [x] upgrade Alchemy as we've hit the apps limit
  - [x] generate API URLs on Alchemy
  - [x] set API URL secrets in Fly.io (see chain suport checklist for instructions)


boudra:
I used Infura in the end because Alchemy doesn't support Avalanche.  @gravityblast I reused the "Gitcoin Production" API keys but we need to ideally organize it better, made a task for it here, I will suggest to be added to next week sprint: https://github.com/gitcoinco/allo-indexer/issues/309

boudra:
Note: Indexing with the default Avalanche RPCs will not work with the current Chainsauce log fetching logic, I implemented a fix but I ended up reverting it here since the larger Chainsauce upgrade will fix this in a safer way.

boudra (approved):

src/config.ts
```diff
@@ -488,7 +488,7 @@ const CHAINS: Chain[] = [
     rpc: rpcUrl
       .default("https://rpc-mumbai.maticvigil.com/")
       .parse(process.env.POLYGON_MUMBAI_RPC_URL),
-    pricesFromTimestamp: Date.UTC(2023, 9, 19, 0, 0, 0),
+    pricesFromTimestamp: Date.UTC(2023, 8, 19, 0, 0, 0),
```
boudra, line 491: @bard I just noticed these changes are in Polygon, should they be here?

bard: Yes, I noticed the date was set in the future there so I fixed it.

vacekj (approved):

PR #315, opened 2023-10-11
Upgrade chainsauce

### LINKED ISSUE #231: rebuild state on startup from event log
Combined with the IPFS cache, this will bring down startup time significantly.

### LINKED ISSUE #310: Upgrade to Chainsauce v2
Chainsauce v2 (https://github.com/boudra/chainsauce/pull/10) brings in a lot of improvements: 

- Event and contract type safety
- Improved RPC fetching logic
- Resumable indexing support 
- Event caching support
- Testability

It should be an easy upgrade.

boudra:
closes #310
closes #231 

Summary of changes:

- Improved log fetching
  - More robust concurrency & retry implementation
  - Fetched sequentially with a simpler way to deal with wide range or too many logs errors
  - Proper log caching, re-indexing will only call the RPC for new blocks
- Automatic types from ABI
  - Contract reads and event parameters are typed from the ABI
- Improved contract reads
  - Reads done at the same block we're indexing, making them deterministic
  - Reads are now cached, reducing RPC requests
- Async IO for JSON storage class
  - Makes the blockchain indexer a better neighbour to other code running in the same process
- Improved watch mode indexing
  - Indexer errors (RPC, event handlers etc.) do not stop watch mode indexing
  - Errors are reported as an event, which we are now logged and sent to Sentry
  - This could potentially fix indexers stopping in production (haven't confirmed this, it's a guess)
- Single event handler functions are supported if we want to split `handleEvent`
- Extract Cache class from Chainsauce, the objective is to get rid of it soon, at the moment only the Prices updater is using it
- Enable the Chainsauce cache in `${CACHE_DIR}/chainsauceCache.db`
- Enable the subscription store to support resumable indexing,  this is stored in the chain data directory as a SQLite database, that directory still being deleted on startup so it still indexes from scratch, but the capability is there when we figure out the best way to do it

I apologize for the big ABI diffs, they're just changing from JSON to TS, I messed up the `git mv` by not committing it.

## Timings

Everything ran with a warm cache.

```
# mainnet
main: npm start 92.64s user 32.44s system 63% cpu 3:16.78 total
feat/upgrade-chainsacue: 81.51s user 19.46s system 102% cpu 1:38.81 total

# optimism
main: 334.98s user 58.26s system 53% cpu 12:21.85 total
feat/upgrade-chainsauce:  48.35s user 96.98s system 65% cpu 3:42.66 total
```

Note: I didn't see such a big difference in timings when deploying this branch to staging.

While checking the snapshot diffs, I found that current `main` is duplicating some votes for the round `0x12BB5bBbFE596dbc489d209299B8302c3300fa40`, it is fixed by this PR, not sure if it's worth investigating further why it happened.


src/cache.ts
```diff
@@ -0,0 +1,82 @@
...
+  private loading: Record<string, Promise<unknown>>;
+  private isDisabled: boolean;
+
+  constructor(dir: string, isDisabled = false) {
+    this.dir = dir;
+    this.loading = {};
+    this.isDisabled = isDisabled;
+
+    if (!this.isDisabled) {
+      mkdirSync(this.dir, { recursive: true });
```
bard, line 17: This will block the entire process. Since it's not possible to make a constructor async, an alternative would be to call the async version lazily at the point where the directory is needed.

src/cache.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+  }
+
+  private key(key: string) {
+    return crypto.createHash("sha256").update(key).digest("hex");
+  }
+
+  private filename(key: string) {
+    return path.join(this.dir, this.key(key));
```
bard, line 26: As someone reading the code, it's a little confusing why a `key` function would be needed here since the key has already been provided. Changing the name slightly for either the function or the parameter might clarify that, e.g.:

```
private filepath(key: string) {
  return path.join(this.dir, this.filenameForKey(key))
}
```

src/cache.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  async get<T>(key: string): Promise<T | undefined> {
+    if (this.isDisabled) {
+      return undefined;
+    }
+
+    const filename = this.filename(key);
+
+    try {
+      return JSON.parse((await fs.readFile(filename)).toString()) as T;
```
bard, line 37: If you pass the encoding parameter to `readFile` (`readFile(filename, 'utf8')`), the `.toString()` should be unnecessary.

src/cache.ts
```diff
@@ -0,0 +1,82 @@
...
+    if (this.isDisabled) {
+      return;
+    }
+
+    const filename = this.filename(key);
+
+    try {
+      await fs.writeFile(filename, JSON.stringify(value));
+    } catch {
+      return undefined;
```
bard, line 53: Swallowing exceptions here might hide serious trouble (e.g. disk out of space errors).

src/cache.ts
```diff
@@ -0,0 +1,82 @@
...
+        return promise;
+      }
+    });
+
+    void this.loading[key].then(() => {
+      delete this.loading[key];
+    });
+
+    return this.loading[key] as Promise<T>;
+  }
```
bard, line 81: The what, why, and how of this function are not immediately clear at least to me; if expressiveness is an objective here, it might be worth rethinking this one, otherwise I'd suggest covering the "what" in a comment.

boudra:
@bard I copied that file as it is from Chainsauce, but I'll fix it up.

bard:
> @bard I copied that file as it is from Chainsauce, but I'll fix it up.

:facepalm: sorry, I didn't realize, and I also missed the fact that this was still marked as draft.

boudra:
> > @bard I copied that file as it is from Chainsauce, but I'll fix it up.
> 
> 🤦 sorry, I didn't realize, and I also missed the fact that this was still marked as draft.

No worries, I was planning on just extracting the class from Chainsauce, leave it as is and hopefully replace it soon with SQLite or cacache but it makes sense to just boy scout it and fix it up now, I'll add some tests for it too.

boudra:
@bard I made some more updates to the Chainsauce interface based on our conversations, the README (https://github.com/boudra/chainsauce) and this PR should provide a good idea of it, but I am still writing up some more docs.

src/config.ts
```diff
@@ -17,8 +16,8 @@ export type Token = {
 };
 
 export type Subscription = {
-  address: string;
-  abi: ethers.ContractInterface;
+  address: `0x${string}`;
```
vacekj, line 19: this can be the `Hex` type from viem.

src/indexer/handleEvent.ts
```diff
@@ -147,84 +126,105 @@ async function handleEvent(
...
       ] = await Promise.all([
-        contract.applicationMetaPtr(),
-        contract.roundMetaPtr(),
-        contract.token(),
-        contract.applicationsStartTime(),
-        contract.applicationsEndTime(),
-        contract.roundStartTime(),
-        contract.roundEndTime(),
-        matchAmountPromise,
+        readContract({
```
vacekj, line 176: viem batches rpc events, but could we speed it up further using multicall?

src/diskCache.ts
```diff
@@ -0,0 +1,80 @@
...
+        .mkdir(this.dir, { recursive: true })
+        .then(() => {
+          return;
+        });
+    }
+    return this.initializePromise;
+  }
+
+  async get<T>(key: string): Promise<T | null> {
+    await this.ensureInitialized();
```
bard, line 33: Nice trick, saving this in my mental notebook.

src/index.ts
```diff
@@ -236,82 +232,121 @@ async function catchupAndWatchChain(
...
+      : null;
+
+    // the subscription store is used to keep track of indexed events for resuming
+    // indexing after a restart, it goes with the chain data
+    const subscriptionStore = createSqliteSubscriptionStore(
+      path.join(CHAIN_DIR_PATH, "subscriptions.db")
+    );
+
+    const indexer = createIndexer({
+      contracts: abis,
```
bard, line 272: If the `contracts` parameter here is about definitions rather than deployments, it might be worth to name that `abis` (or something similarly unambiguous), too?

src/index.ts
```diff
@@ -236,82 +232,121 @@ async function catchupAndWatchChain(
...
+          url: config.chain.rpc,
+        }),
+      },
+      context: eventHandlerContext,
+      subscriptionStore,
+      cache: chainsauceCache,
+      logLevel: "trace",
+      logger: (level, data: unknown, message?: string) => {
+        if (level === "error") {
+          indexerLogger.error(data, message);
```
bard, line 289: I'm not sure whether pino methods accepts multiple arguments. If they don't, then this should work:

```ts
indexerLogger.error({ msg: message, data })
```

src/indexer/handlers/applicationMetaPtrUpdated.ts
```diff
@@ -1,25 +1,27 @@
...
+  const metaPtr = event.params.newMetaPtr.pointer;
   const metadata = await ipfsGet<Round["applicationMetadata"]>(metaPtr);
 
   await db.collection<Round>("rounds").updateById(id, (round) => {
     return {
       ...round,
-      applicationMetaPtr: event.args.newMetaPtr.pointer,
-      updatedAtBlock: event.blockNumber,
+      applicationMetaPtr: event.params.newMetaPtr.pointer,
+      updatedAtBlock: Number(event.blockNumber),
```
bard, line 24: `BigInt`? (since I saw you fixed it elsewhere)

bard (dismissed):

src/index.ts
boudra: yeah you're right, naming it `abis` will make the separation clearer 👍

src/index.ts
boudra: this seems is working as intended at the moment, but I'll double check the arguments passed

boudra: changed to use the object interface, it's clearer as well

src/config.ts
boudra: fixed

src/index.ts
boudra: in the end I kept the name `contracts` for this PR because it's not a simple rename here and needs more thought , because the name here affects other parts of the interface, I was thinking this could be an option but I am still not 100% sure:

```
import RoundImplementationV2  from "./abis";

const indexer = createIndexer({
    abis: {
       RoundImplementation
    },
   ...
});

indexer.subscribeToContract({
   abiName: "RoundImplementation",
   address: "0x...",
});

indexer.readContract({
    abiName: "RoundImplementation",
    address: "0x...",
    functionName: "roundMetaPtr"
});
```

so the way the interface is now the ABI is abstracted away after indexer creation and you only deal with "contracts":

1. you define your contracts (which are just the ABIs)
2. you subscribe to a deployed contract referencing the contract name
3. event handlers read from deployed contracts referencing the contract name

renaming contracts to ABIs:

1. you define your ABIs
2. you subscribe to a deployed contract referencing the ABI name
3. event handlers read from deployed contracts referencing the ABI name

I feel like calling them ABIs is the right name in the `createIndexer` call but down the line you're dealing with contracts and the ABI is an "implementation detail".

boudra:
Ready to merge now, other work like resumable indexing and indexing status endpoint depend on this one.

vacekj (approved):

src/index.ts
boudra: resolving the conversation to merge, but open to revisit this later

PR #320, opened 2023-10-13
fix: big numbers scientific notation problem

### LINKED ISSUE #319: fix: number.toString issue
when converting very big numbers to strings using `.toString()`, js uses scientific notation above a certain treshold (`123e12`). this breaks parsing this string back into BigInts. we avoid this by first converting this number to bigints, then converting to string, then converting back to bigint.

vacekj:
fixes #319


src/calculator/index.ts
```diff
@@ -321,24 +323,27 @@ export default class Calculator {
 
         /*TODO: take into account round matching token decimals instead of default 18 */
         const amountRoundToken =
-          (Number(amountUSD) / conversionRateRoundToken.price) *
-          Math.pow(10, 18);
+          (Number(amountUSD) / conversionRateRoundToken.price) * 10e18;
```
gravityblast, line 326: you can use `.toLocaleString("fullwide", { useGrouping: false })` for that. 
otherwise we should convert the numbers before the multiplication that make them big in case they become too big for numbers

src/calculator/index.ts
```diff
@@ -321,24 +323,27 @@ export default class Calculator {
...
 
         return {
           ...vote,
           amount: vote.amount.toString(),
-          amountRoundToken: Math.floor(amountRoundToken).toString(),
+          amountRoundToken: BigInt(Math.floor(amountRoundToken)).toString(),
```
gravityblast, line 331: you can use .toLocaleString("fullwide", { useGrouping: false }) for that.
otherwise we should convert the numbers before the multiplication that make them big in case they become too big for numbers

vacekj: Relying on BigInt's toString feels better than relying on a toLocaleString, but I'll change it. Let's just get this merged asap, this is a blocker for testing and shipping matching estimates.

vacekj: I'm gonna make the call to rely on the standard BigInt toString instead of the toLocaleString, because it's clearer what the code is doing. requesting approval @boudra @gravityblast @bard @bhargavaparoksham

bard: +1 from me.

src/calculator/votes.ts
```diff
@@ -65,6 +69,14 @@ export async function getVotesWithCoefficients(
     const passportScore = await passportProvider.getScoreByAddress(voter);
     let passportCheckPassed = false;
 
+    if (
+      options.bypassPassportAddresses
+        ?.map((address) => address.toLowerCase())
```
bard, line 74: Thinking aloud / out of scope for this PR: at some point we might want to use a type to carry information about address validity along with the address itself, since otherwise the only defense against a nasty case bug is our memory.

bard (dismissed):

src/calculator/index.ts
boudra: @vacekj looking at this code closer, I think there are a couple of places where we could lose precision:

1. `formatUnits(vote.amount, 18)` converting the bigint to a number string here and parsing it again as a number is unnecessarily risky
2. `const amountRoundToken = (Number(amountUSD) / conversionRateRoundToken.price) * 10e18;` this multiplication will easily result a higher number than `Number.MAX_SAFE_INTEGER` (9.07*10e15), leading to possible loss of precision
3. `Math.floor(amountRoundToken)` is also a potential loss of precision
4. We're not taking into account the token decimals, this would be fine if we didn't have support USDC which has different decimals than the rest

not sure if you saw how we're doing the conversion here: https://github.com/gitcoinco/allo-indexer/blob/main/src/prices/provider.ts#L104

in that calculation, we truncate the USD price to 8 places and then do all the math in big integers, it also takes into account the token decimals

this could all still be working fine, but if we already have a tested calculation in place, I would try to use that instead of coming up with new less precise calculation, that calculation can be split up into it's own function if needed

bard: @vacekj @boudra would the above be reason enough to have a go-to `token-math.ts` module (or something like that) where this sort of calculations is centralized and, more importantly, tested against edge cases?

boudra: @bard This type of math can get complex easily so I think it's the way to go. I would already do it in this PR, off the top of my head I think all we would have to do is split up the logic that already exists into a pure functions in `tokenMath.ts` and implement some tests for it, I'd be more than happy to help here @vacekj since I implemented those first calculations.

boudra: I also noticed that those calculations have a lot of math in one line expressions so they might be hard to read, again happy to help here!

boudra:
This is ready for a review @gitcoinco/grantsstack-engineering 

fixed a couple of problems and extracted the token conversion functions

snapshot tests reported no difference 👍

boudra (approved):
LGTM 🙃

src/calculator/votes.ts
vacekj: Very relevant thread: https://github.com/ethers-io/ethers.js/discussions/3089

bhargavaparoksham (approved):

src/calculator/votes.ts
boudra: I like the second option, the internal value can still technically be messed with but the naming will make it look wrong, it fixes the duck typing issue and there's no Typescript magic to wrap your head around.

src/calculator/index.ts
```diff
@@ -297,48 +300,63 @@ export default class Calculator {
...
-      potentialVotes.map(async (vote) => [
-        vote.token,
-        await this.priceProvider.getUSDConversionRate(this.chainId, vote.token),
-      ])
-    );
-    const prices = Object.fromEntries(conversionRates) as Record<
-      string,
-      Price & { decimals: number }
-    >;
+    const prices: Record<string, PriceWithDecimals> = {};
```
bard, line 311: Nit: it can be inferred by the code immediately below, but if you want the reader to save that time, this could be renamed to `usdConversionRateByTokenAddress` or something like that (assuming I'm inferring correctly here).

src/calculator/index.ts
```diff
@@ -297,48 +300,63 @@ export default class Calculator {
...
-      ...potentialVotesAugmented,
-    ]);
+      const amountUSD = convertTokenToFiat({
+        tokenAmount: vote.amount,
+        tokenDecimals: tokenPrice.decimals,
+        price: tokenPrice.price,
+        priceDecimals: 10,
+      });
+
+      const amountRoundToken = convertFiatToToken({
```
bard, line 336: Is `round` here is meant as in `Math.round` rather than "funding round"? If so, I'd suggest changing it to something able to remove all doubts, like `roundedTokenAmount` (key point being `roundED`).

src/calculator/index.ts
```diff
@@ -297,48 +300,63 @@ export default class Calculator {
...
-      })
-    );
+    const potentialVotesAugmented: Vote[] = potentialVotes.map((vote) => {
+      const tokenPrice = prices[vote.token];
 
-    const potentialResults = await this._calculate([
-      ...votes,
-      ...potentialVotesAugmented,
-    ]);
+      const amountUSD = convertTokenToFiat({
```
bard, line 329: This diff by the way is a nice example of ensuring a consistent abstraction level within a function body as mentioned recently by @lucianHymer

src/tokenMath.ts
```diff
@@ -0,0 +1,47 @@
...
+ * @param args.tokenAmount The token amount
+ * @param args.tokenDecimals The token decimals
+ * @param args.price The price of the token
+ * @param args.priceDecimals The decimal places to use for the price, defaults to 8, any decimals beyond this will be truncated
+ * @returns The currency amount
+ * */
+export function convertTokenToFiat(args: {
+  tokenAmount: bigint;
+  tokenDecimals: number;
+  price: number;
```
bard, line 11: If `price` is the price of the token, maybe worth naming it `tokenPrice` so it's also immediately clear at call sites just by looking at the invocation?

bard (dismissed):
Nice one :+1:

src/calculator/votes.ts
boudra: we could go a step further an name it `__addressDoNotAccessDirectly`, make it look really wrong!

boudra: moved conversation here 👉 https://github.com/gitcoinco/allo-indexer/issues/324

boudra:
Conflicts resolved and feedback applied.

bard (approved):

boudra:
This is working with the Grants stack PR as well 🚀 

Thanks everyone for the reviews, and great work @vacekj 💪 

Merging!

PR #321, opened 2023-10-17
Proportional matches

gravityblast:
closes https://github.com/gitcoinco/grants-stack/issues/2387

* [x] remove custom `passportThreshold`
* [x] add new score/percentage range options
* [x] set calculations coefficient based on proportional matching setting

### DIFF
--- a/src/calculator/index.ts
+++ b/src/calculator/index.ts
@@ -16,6 +16,7 @@ import {
 import { PotentialVote } from "../http/api/v1/matches.js";
 import { Price } from "../prices/common.js";
 import { formatUnits, zeroAddress } from "viem";
+import { ProportionalMatchOptions } from "./options.js";
 
 export {
   CalculatorError,
@@ -165,16 +166,21 @@ export default class Calculator {
     });
   }
 
-  async calculate(): Promise<Array<AugmentedResult>> {
+  async calculate(
+    pmOptions: ProportionalMatchOptions
+  ): Promise<Array<AugmentedResult>> {
     const votes = await this.parseJSONFile<Vote>(
       "votes",
       `${this.chainId}/rounds/${this.roundId}/votes.json`
     );
 
-    return this._calculate(votes);
+    return this._calculate(pmOptions, votes);
   }
 
-  private async _calculate(votes: Vote[]): Promise<Array<AugmentedResult>> {
+  private async _calculate(
+    pmOptions: ProportionalMatchOptions,
+    votes: Vote[]
+  ): Promise<Array<AugmentedResult>> {
     const applications = await this.parseJSONFile<Application>(
       "applications",
       `${this.chainId}/rounds/${this.roundId}/applications.json`
@@ -224,6 +230,7 @@ export default class Calculator {
     }
 
     const votesWithCoefficients = await getVotesWithCoefficients(
+      pmOptions,
       this.chain,
       round,
       applications,
@@ -284,6 +291,7 @@ export default class Calculator {
    * @param roundId
    */
   async estimateMatching(
+    pmOptions: ProportionalMatchOptions,
     potentialVotes: PotentialVote[],
     roundId: string
   ): Promise<MatchingEstimateResult[]> {
@@ -299,7 +307,7 @@ export default class Calculator {
 
     const round = rounds.find((round) => round.id === this.roundId) as Round;
 
-    const currentResults = await this._calculate(votes);
+    const currentResults = await this._calculate(pmOptions, votes);
     const conversionRates = await Promise.all(
       potentialVotes.map(async (vote) => [
         vote.token,
@@ -335,7 +343,7 @@ export default class Calculator {
       })
     );
 
-    const potentialResults = await this._calculate([
+    const potentialResults = await this._calculate(pmOptions, [
       ...votes,
       ...potentialVotesAugmented,
     ]);
--- a/src/calculator/index.ts
+++ b/src/calculator/index.ts
@@ -103,6 +103,7 @@ export type CalculatorOptions = {
   ignoreSaturation?: boolean;
   overrides: Overrides;
   chain: Chain;
+  proportionalMatchOptions: ProportionalMatchOptions;
 };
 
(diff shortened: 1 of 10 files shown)


src/calculator/index.ts
```diff
@@ -98,11 +99,11 @@ export type CalculatorOptions = {
   roundId: string;
   minimumAmountUSD?: number;
   matchingCapAmount?: bigint;
-  passportThreshold?: number;
   enablePassport?: boolean;
   ignoreSaturation?: boolean;
   overrides: Overrides;
   chain: Chain;
+  proportionalMatchOptions: ProportionalMatchOptions;
```
boudra, line 106: Everything in CalculatorOptions is already an "option" so there might not be a need to have an Options suffix here, like the other options in it.

src/calculator/votes.ts
```diff
@@ -17,8 +18,8 @@ export async function getVotesWithCoefficients(
   options: {
     minimumAmountUSD?: number;
     enablePassport?: boolean;
-    passportThreshold?: number;
-  }
+  },
+  pmOptions: ProportionalMatchOptions
```
boudra, line 22:
```suggestion
  proportionalMatchOptions: ProportionalMatchOptions
```

I would suggest to have the full name here

boudra:
There are some linter errors

src/http/api/v1/exports.ts
```diff
@@ -90,7 +91,8 @@ export const createHandler = (config: HttpApiConfig): express.Router => {
       applications,
       votes,
       config.passportProvider,
-      {}
+      {},
+      defaultProportionalMatchOptions
```
bard, line 95: Here we're grabbing something from the calculator module (`import { defaultProportionalMatchOptions } from '../../../calculator/options.js' `) and passing it back to the calculator, unmodified. Since it's a default, the calculator shouldn't need to be told about it, so I'd opt for leaving it `undefined`, letting `getVotesWithCoefficient` fill in the blank, and dropping a module dependency.

That said, `getVotesWithCoefficient` could really use a move from positional arguments to a data object here.

src/http/api/v1/matches.ts
```diff
@@ -94,13 +94,11 @@ export const createHandler = (config: HttpApiConfig): express.Router => {
...
         ? BigInt(matchingCapAmount)
         : undefined,
-      passportThreshold: passportThreshold
-        ? Number(passportThreshold)
-        : undefined,
       enablePassport: enablePassport,
       ignoreSaturation: ignoreSaturation,
       overrides,
       chain: chainConfig,
+      proportionalMatch: defaultProportionalMatchOptions,
```
bard, line 101: Same as above.

src/test/calculator/proportionalMatch.test.ts
```diff
@@ -0,0 +1,196 @@
...
+
+  return { vote, score };
+}
+
+describe("getVotesWithCoefficients", () => {
+  let votes: Vote[] = [];
+  let scores: PassportScore[] = [];
+  let fakePassportProvider: FakePassportProvider;
+
+  beforeAll(() => {
```
bard, line 139: If you want to get fancy / extra tidy here, vitest has a "test context" that works great for separating fixture code from test code: https://vitest.dev/guide/test-context.html

There's an example of that in the passport provider tests.

bard (dismissed):

src/http/api/v1/exports.ts
gravityblast: @bard good ideas, I removed positional args and set the proportionalMatchingOptions optional using the default by..default

src/calculator/votes.ts
```diff
@@ -30,18 +35,18 @@ export async function getVotesWithCoefficients(
   );
 
   const enablePassport =
-    options.enablePassport ??
-    round?.metadata?.quadraticFundingConfig?.sybilDefense ??
+    args.options.enablePassport ??
```
bard, line 38: Just FYI you can still enjoy named parameters here by deconstruction if you wish, as in:

```ts
export  async function getVotesWithCoefficients(
  { applications, options, rounds }: GetVotesWithCoefficientsArgs
): ... {
  // ...
  const enablePassport = options.enablePassport ?? ...
```

(But `args` is just as fine.)

bard (dismissed):

src/http/api/v1/matches.ts
gravityblast: done

src/calculator/votes.ts
gravityblast: @bard yeah that was the initial implementation but it was very noisy, so I moved to args

gravityblast:
@bard I extended the test with fixtures like you proposed

src/calculator/votes.ts
```diff
@@ -102,6 +100,31 @@ export async function getVotesWithCoefficients(
   return (await Promise.all(votePromises)).flat();
 }
 
+export function scoreToCoefficient(
```
boudra, line 103:
```suggestion
function scoreToCoefficient(
```

it's an internal function right? not seeing usage outside

gravityblast: @boudra yes now that it's there it makes sense I forgot it, thank you!

boudra (dismissed):

src/test/calculator/proportionalMatch.test.ts
```diff
@@ -0,0 +1,54 @@
...
+  beforeAll(() => {});
+
+  describe("should update the amount proportionally based on the passport score", () => {
+    test("returns votes with amounts updated proportionally based on passport score", async ({
+      chain,
+      round,
+      applications,
+      data,
+    }) => {
+      fakePassportProvider = new FakePassportProvider(data.scores);
```
boudra, line 20:
```suggestion
      const fakePassportProvider = new FakePassportProvider(data.scores);
```

vacekj (dismissed):

gravityblast:
@boudra pushed a fix to your last comment thank you

gravityblast:
@boudra @vacekj

boudra (approved):

PR #326, opened 2023-10-19
fix: streamline variable data dir

### LINKED ISSUE #318: add `.var` in gitignore
if we add a `.var` entry in gitignore then for development we can have in .env:

```
STORAGE_DIR=./.var/storage/data
CACHE_DIR=./.var/storage/cache
```

And also passport_scores.leveldb won't appear in the worktree. So we can ignore/exclude just .var everywhere, such as `vitest.config.ts`

boudra:
closes #318 

streamlines all variable data directories inside `.var`

if your data lives under `./storage`, do:

```
mkdir .var
mv storage .var
```


src/index.ts
```diff
@@ -97,13 +97,13 @@ async function main(): Promise<void> {
     ]);
 
     const httpApi = createHttpApi({
-      storageDir: config.storageDir,
+      storageDir: config.chainDataDir,
```
boudra, line 100: is this confusing? the storage dir inside the http api context is the chain data directory, but it could be confused with the root storage dir because of the name

bard: Maybe it could be:

```ts
const httpApi = createHttpApi({
  httpRootDir: config.chainDataDir
```

So as to keep "storage dir" to one meaning?

bard (approved):

src/index.ts
boudra: updated :v:

gravityblast (approved):

PR #329, opened 2023-10-19
Fix block cache

boudra:
the block cache was not being initialized, but it wasn't used either so it never threw any errors 🤦 

I didn't test properly after addressing PR feedback

hopefully we can get to this one soon: https://github.com/gitcoinco/allo-indexer/issues/328


bard (approved):
Looks good. I also had missed this in review.

PR #331, opened 2023-10-23
fix: create chain data directory if it doesn't exist

boudra:
the chain data directory is not created before starting the passport service, leading to an error when first writing `passport_scores.json`, the directory is eventually created so the problem doesn't happen after a restart

note: `chainDataDir` is actually `${storageDir}/chainData` but as far as the passport service is concerned they could be two totally different directories, so we create both of them


vacekj (approved):
:shipit:

Issue #316, opened 2023-10-12
implement resumable indexing


boudra:
We might not need to do this if the depoyments become reliable and fast enough.

from @bard:

>  Separate instances make it tricky to do anything like setting a flag in a db or on the filesystem. Maybe the build tag could be used in conjunction with Fly secrets, i.e. fly secret set REINDEX_AT_REV=0a586b --staged and then in the code if (config.buildTag === config.reindexAtStage) { reindex() }

boudra:
@bard what do you think about this approach:

1. We introduce the concept of a "data version number", we could put it in the codebase as a hard coded value or as an environment variable that we modify directly in the fly.toml, so this version number is coupled to the codebase, when we make a schema change we increment this number
2. Each instance tracks the currently indexed version in a file, if the version indexed in disk is different from the one that we're currently running, we delete the chain directory and reindex

Basically the same approach you proposed but we track the version with the code instead of in a secret, this way we can review it in PRs and there's no "manual step", there's still the chance that we forget to change this number but I think it will be hard to come up with a solution where that is not an issue.

boudra:
Also from our other conversation in Notion, I agree that we can skip doing SQL migrations and just use whatever system we come up here to reindex instead of migrating since all data we have is always sourced from elsewhere.

bard:
> @bard what do you think about this approach:

Yeah, this sounds much better, is a battle-tested approach and keeps a paper trail in git.

Issue #314, opened 2023-10-10
replace JSON files with Postgres database


boudra:
We're storing the data in JSON files, as the data gets larger this is becoming more of a bottleneck to write to, it is also difficult to query, making it hard to support more advanced features in Grants Stack, like the new homepage.

I suggest we use https://kysely.dev/ as a database abstraction.

vacekj:
"Awesome, hell yeah, perfect, let's go" :shipit:

Issue #173, opened 2023-07-20
replace static files with API endpoints


bard:
We're currently exposing JSON files as part of the public API. We should instead provide the same information through API handlers (in a backward-compatible way).

Related: #268

Issue #114, opened 2023-06-26
unit test event handlers


boudra:
The switch is big and it's hard to follow. We can put event handlers into functions and calling those functions from the switch.

See `matchAmountUpdated` as an example.

bard:
An important goal of this is to improve testability by having smaller, well-defined units of functionality.

bard:
poc: https://github.com/gitcoinco/allo-indexer/pull/161

PR #345, opened 2023-11-17
PostgreSQL + GraphQL

boudra:
# What is in this PR

- [x] Postgres as a data backend for chain data via `Kysely` and `pg`
- [x] `handleEvent` refactor to return mutations instead of executing them directly
- [x] Renaming of entities and field names (for example, `votes => totalDonationsCount`, `metaPtr => metadataCid`)
- [x] Automatic GraphQL and GraphiQL endpoints from Postgres via Postgraphile
- [x] Resumeable indexing through Chainsauce (brings HTTP startup time to minutes)
- [x] Backwards compatible API and `DataProvider` serving legacy JSON files, to be removed later
- [x] Tests for `handleEvent` & remove snapshot test
- [x] Postgres as a data backend for prices

closes https://github.com/gitcoinco/grants-stack-indexer/issues/316
closes https://github.com/gitcoinco/grants-stack-indexer/issues/316
closes https://github.com/gitcoinco/grants-stack-indexer/issues/314
closes https://github.com/gitcoinco/grants-stack-indexer/issues/173
closes https://github.com/gitcoinco/grants-stack-indexer/issues/114

#  Why Postgres, what happened to SQLite?

I made a quick spike to see how SQLite would work: https://github.com/gitcoinco/grants-stack-indexer/pull/338

In the end I decided to try a spike with Postgres because:

- The effort to get a first working version was about the same with the architecture implemented in this PR, you could argue that it was easier with Postgres as the `SCHEMA` and big integer support played a big role in getting this done
- SQLite actually worked really well and indexing was way faster than JSON or Postgres, but it is clear that the data accessibility outside of us making our own APIs is not as good, with Postgres it's as easy as spinning off a read replica and give read access to the data team
- Postgres has the upper hand in concurrency control, in SQLite we can only have a single transaction per connection, and only one connection can write at once
- Postgres has very good support for automatic APIs: PostgREST, Postgraphile, Hasura, pg_graphql. Which can cut down on development time and maintenance (or at least that's my hope)

Given these reasons and the fact that everyone already seemed on board, it was an easy choice.

I think SQLite can play a big role in having a portable cache if we don't want to burden Postgres with the caches, I can imagine us having some kind of `cache.db` where we have (IPFS data, event logs, contract calls, coingecko caches) and being able to move it around easily between environments. Like run `npm run fetch-cache`, and you're ready to index fully offline, for example.

# Why GraphQL?

This one was probably more of a surprise as we haven't really talked about it, this came up after deciding to try out Postgres, I initially had considered these options:

- Managed Supabase
    - Advantages: Managed Postgres + automatic PostgREST + GraphQL APIs, I have been using it for a couple of projects and I have been very impressed with their offering
    - Disadvantages: It comes bundled with a bunch of a lot more stuff which we don't need and there's a risk to be locked in to a degree, even if we can self host it
- Postgres + pg_graph
    - Advantages: pg_graph is built by Supabase and powers their automatic GraphQL APIs
    - Disadvantage: We have to start messing with Postgres extensions, and most managed providers do not have support for custom extensions, we would have to manage our own database
- Postgres + PostgREST (my personal favourite)
    - Advantages: Battle tested automatic REST API, paired with Supabase's `potgrest-js` https://github.com/supabase/postgrest-js would make it a lot more flexible than a GraphQL API and we get a client for free
    - Disadvantages: It complicates our infrastrcutrue since we'd have to deploy it separately, considering we have staging and production to think about.
- Fly Postgres + Postgraphile
    - Advantages: Postgraphile is a Typescript library, it can be bundled with our own API, making deployment and maintenance much much easier, it seems to have been around for a long time too and read good things about it, it has a Pro version at $25/mo which gives us support from the developer. This also allows us to use vanilla Postgres and we are not locked into Fly Postgres. Using Fly Postgres has the advantage that we can put the database in the same data center as the worker reducing latency and everything is integrated in the same CLI.
    - Disadvantages: Fly Postgres is technically [not "managed postgres"](https://fly.io/docs/postgres/) altough it's pretty close, it suports the usual features: scaling, read replicas, failover, backups etc. (and in our case the database is actually a derived data form other sources so I don't think it's a big issue).

# Architecture overiew

The main objective of this PR is to unblock work happening in the front end which needs a data API to avoid making hundreds of requests to various sources for a simple page load. It is also apparent that we're making it harder than it should be for the data team to query the data, as they have to build ETL pipelines on top of the JSON data.

Therefore, I have prioritized keeping the scope small to accomplish this objective, altough I have used this chance to tackle some low hanging fruits and I believe this is a very good step in the right direction.

It is clear that we are due another design review as most of the big issues we discussed on our last indexer roadmap dsicussions are now fixed, and with the introducion of Postgres we have new challenges to think about.

At the moment even though we are using a server database, each instance is indexing to their own schema:

- Fly app
    - Fly instance A -> Postgres (schema: `chain_data_A_1`)
    - Fly instance B -> Postgres (schema: `chain_data_B_1`)

The number at the end of the schema name (`chain_data_A_1`) is for versioning which is required for resumable indexing, which we need if we don't want to reindex on every deployment, at the moment data insertion in Postgres is unoptimized (we insert votes one by one), which increases total indexing time up to 5 hours, there are things we can do about it, like batched inserting, but I have avoided doing any premature optimization in this PR and want to make it work correctly first.

Resumable indexing makes deployments very fast (6 minutes for the two staging instances).

Each deployment will pick up the same schema that was already indexed and continue, and if we make any breaking change, we can increment the version number in the codebase and the indexer will automatically re-index from scratch, I suspect we won't do this a lot.

The reason for having each instance use their own schema, is because making the change to have instances read and/or write to the same schema would require some coordination between them or a complete re-architecture, and I think whichever way we decide to go, it would have to be done separately, here are two ideas that came up while thinking about this to get the discussion started:

**Homogenous deployment**

All instances deploy the same codebase (indexer + API), this is what we do now, it works because each instance manages their own database and is completely independent of the others, but sharing a database will require coordination, I was thinking we could use a simple database level lock to signal writing, it could work this way:
    - Instance A starts up, tries to acquire database level dock (potentially through Postgres advisory locks), if acquired, it indexes
    - Instance B starts up, tries to acquire the lock and fails, doesn't run the indexer, only the API and polls periodically for the lock to be released
    - Instance A stops for whatever reason, instance A acquires the lock and starts indexing

This way we shift the complexity from the orchestration to the applictation, all instances are the same, but only one can index.

This can be problematic as the API and the indexer have different performance considearions.

**Re-design infrastructure**

we could deploy a single indexer instance (writer) per environment and multiple web instances (readers).

It wouldn't be as straightforward as your typical worker as there is a dependency here, the web instances should not serve traffic until the indexer is up to date, so there might be some coordination that needs to be done here, either on the orchestrator level or the application level.

# Renames

I also taken advantage of this and renamed some of the entities and column names, please review the schema and make any suggestions!


src/database/index.ts
```diff
@@ -0,0 +1,344 @@
...
+  rounds: RoundTable;
+  applications: ApplicationTable;
+  donations: DonationTable;
+}
+
+type KyselyDb = Kysely<Tables>;
+
+type ChainId = number;
+
+export type Mutation =
```
boudra, line 36: @bard I have basically copied the query implementation of the data client here, I have split off mutation into their own type and method, so we have `mutate` and `query`. My thinking was if we make all mutations are all fire and forget, it gives us more flexibility when composing those mutations (like batching etc.) so I split them up to make it clear they work differently, but they can be technically unified within the `query` type, renaming to something like `operation` for example. Would like to hear your thoughts here

vacekj:
Would we be down to explore Supabase more? I'm curious what the true cost (developer time + $$$) of a migration from Supabase to self-hosted Supabase would be. I also have extensive exp. with Supabase and my experience has been great so far.

I feel like it would save us a lot of effort in managing infra, and speed up development even more, while still having that escape hatch of self-hosting the whole thing, that we can use at any time.

boudra:
I forgot to add something else, i have  experimented with both Drizzle and Kysely, ended up with Kysely because:

- Drizzle didn't have support for dynamic schemas in queries
- Drizzle doesn't have any sort of DDL support, but it forces you to define your schema in Typescript and generate SQL migrations from them, felt a bit weird because I'd do one or the other for me personally.

For the actual query building interface I did like Dirzzle a lot more but it felt very new still, with missing documentation and features.

With Kysely we're creating the schema right in Typescript and when the API doesn't support something it's very easy to use raw SQL.

We can use raw SQL for all this stuff but I have enjoyed having Typescript suggestions for queries and if you know SQL it won't be very different.

PR #348, opened 2023-11-17
Write JSON in chunks

boudra:
actual change here: https://github.com/gitcoinco/chainsauce/compare/b741d3289e7caec8070328d9dd343f9a0a28c289...53bb819bb5fa7fb6d6541ad8cce7151e1895c4df

couldn't use createWriteStream in the end because it is async, and turning that JSON class async is not straightforward, so ended up doing something similar with `openSync + writeSync`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2453,7 +2453,7 @@
     },
     "node_modules/chainsauce": {
       "version": "1.0.19",
-      "resolved": "git+ssh://git@github.com/gitcoinco/chainsauce.git#32cd95f7f984331d87b8fa21292059cce69f412c",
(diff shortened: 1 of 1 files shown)


bard:
Ah... might be risky to do this sync though as it replaces one syscall per set with one syscall per vote. It's a lot of blocking.

boudra:
@bard I agree, I like your idea of writing the file in bigger chunks (splitting in half or more)

I haven't been seeing any out of memory errors and the usage seems stable, I will get an implementation ready if I get a chance, just in case as we're still halfway through the voting period and things could change, but for now we're doing OK.

PR #360, opened 2023-11-21
Implement production tweaks done directly in Fly

boudra:
I had made these tweaks directly in Fly, this checks them into the config.

In production the hard limit is at 40, I want to test increasing to 60. Max old size is set to 8GB, this sets it at 7GB.

These changes were deployed evening 19th, occurrences of reaching hard connection limit errors and out of memory errors reduced significantly:

Screenshot 2023-11-21 at 13 53 17
Screenshot 2023-11-21 at 13 53 49

### DIFF
--- a/fly.production.toml
+++ b/fly.production.toml
@@ -38,7 +38,7 @@ kill_timeout = 5
 
   [services.concurrency]
     type = "connections"
-    hard_limit = 25
+    hard_limit = 40
     soft_limit = 20
 
   [[services.tcp_checks]]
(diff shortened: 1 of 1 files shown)


bard (approved):
Connection limit aside, did you notice any effect from the GC tweak?

boudra:
> Connection limit aside, did you notice any effect from the GC tweak?

Yes, I had noticed around 4 out of memory errors last Sunday, 0 out of memory errors have happened since then and I also noticed the memory now sometimes spiking over 2GB and then going back, which leads me to believe Node is using more memory available to it instead of crashing.

Screenshot 2023-11-21 at 14 28 38

bard:
Indeed. Might be better to wait to have a rate limiter in front of that endpoint first.

PR #366, opened 2023-11-30
http request logger

boudra:


src/http/app.ts
```diff
@@ -34,7 +35,10 @@ interface HttpApi {
 export const createHttpApi = (config: HttpApiConfig): HttpApi => {
   const app = express();
 
+  app.set("trust proxy", true);
```
bard, line 38: @boudra couldn't find a better solution here to get actual IPs than allowing all proxies, because we don't have a fixed IP for Fly's load balancer. Unfortunately this means that the IP can be spoofed. Good to keep in mind.

bard (approved):

PR #370, opened 2023-12-08
Add base support to indexer

bhargavaparoksham:
Add base support to indexer

ref: https://github.com/gitcoinco/grants-stack/issues/2691
contracts: https://github.com/allo-protocol/allo-contracts/blob/main/docs/CHAINS.md#base-mainnet

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -503,7 +503,7 @@ const CHAINS: Chain[] = [
     name: "base",
     rpc: rpcUrl
       .default("https://mainnet.base.org/")
-      .parse(process.env.OPTIMISM_RPC_URL),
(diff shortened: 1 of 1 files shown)


src/config.ts
```diff
@@ -498,6 +498,50 @@ const CHAINS: Chain[] = [
       },
     ],
   },
+  {
+    id: 8453,
+    name: "base",
+    rpc: rpcUrl
+      .default("https://mainnet.base.org/")
+      .parse(process.env.OPTIMISM_RPC_URL),
```
boudra, line 506: shouldn't this be BASE_RPC_URL?

PR #384, opened 2024-01-17
index programs and program roles on Allo V1

gravityblast:


src/indexer/allo/v1/roleRevoked.ts
```diff
@@ -0,0 +1,71 @@
...
+    "RoleRevoked"
+  >
+): Promise<Changeset[]> {
+  const {
+    chainId,
+    event,
+    context: { db, logger },
+  } = args;
+
+  switch (args.event.contractName) {
```
boudra, line 20: maybe this could be an early return to reduce number of blocks here? does this make sense:

```
if(args.event.contractName !== "AlloV1/ProgramFactory/V1") {
   return [];
}

// event handler here
```

vacekj (approved):
+1 to Mo's comments, otherwise LGTM.

src/indexer/allo/v1/roleRevoked.ts
gravityblast: @boudra I'm going to add more contracts now so the switch should make more sense

PR #408, opened 2024-02-04
override projects metadata with empty object in case it's not an object

gravityblast:
depends on #407


boudra (approved):

src/indexer/allo/v2/handleEvent.ts
```diff
@@ -48,7 +48,14 @@ export async function handleEvent(
     case "ProfileCreated": {
       const profileId = event.params.profileId;
       const metadataCid = event.params.metadata.pointer;
-      const metadata = await ipfsGet<ProjectTable["metadata"]>(metadataCid);
+      let metadata = await ipfsGet<ProjectTable["metadata"]>(metadataCid);
+
+      // FIXME: update this when we validate all the metadata
+      // for projects and rounds
+      if (Array.isArray(metadata) || typeof metadata !== "object") {
```
boudra, line 55: only thing to be careful here is that `typeof null === "object"` is true

PR #428, opened 2024-02-14
feat: add canonical handling

thelostone-mc:
#### When creating a linked Profile and the canonical already exists

```
{
  projects(
    filter: {id: {equalTo: "0xbfe075166257fda5b915e3dda8be67b44b2db34ddf476841b1e7aa33a5e13db2"}}
  ) {
    id
    chainId
    projectType
    metadata
  }
}
```

image

#### When attempting to create a linked Profile BUT the canonical does not exist

image

Fixes https://github.com/gitcoinco/grants-stack-indexer/issues/427

### DIFF
--- a/src/indexer/allo/v2/handleEvent.ts
+++ b/src/indexer/allo/v2/handleEvent.ts
@@ -29,8 +29,8 @@ function generateRoundRoles(poolId: bigint) {
 function getProjectType(metadata: object) {
   const linkedProjectMetadata = z.object({
     canonical: z.object({
-      chainId: z.number(),
       registryAddress: z.string(),
+      chainId: z.coerce.number(),
     }),
   });
 
@@ -392,6 +392,7 @@ export async function handleEvent(
     case "ProfileMetadataUpdated": {
       const metadataCid = event.params.metadata.pointer;
       const metadata = await ipfsGet<ProjectTable["metadata"]>(metadataCid);
+
       return [
         {
           type: "UpdateProject",
--- a/src/indexer/allo/v2/handleEvent.ts
+++ b/src/indexer/allo/v2/handleEvent.ts
@@ -29,7 +29,7 @@ function generateRoundRoles(poolId: bigint) {
 function getProjectType(metadata: object) {
   const linkedProjectMetadata = z.object({
     canonical: z.object({
-      chainId: z.string(),
+      chainId: z.number(),
       registryAddress: z.string(),
     }),
   });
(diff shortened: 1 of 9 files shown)


src/indexer/allo/v2/handleEvent.ts
```diff
@@ -19,6 +20,20 @@ function generateRoundRoles(poolId: bigint) {
...
+  const canonicalProject = (metadata as Record<string, unknown>)["canonical"] as CanonicalProject;
+
+  if (!canonicalProject) {
+    return null;
+  }
+
+  const canonicalRegistryAddress = canonicalProject.registryAddress;
+  const canonicalChainId = canonicalProject.chainId;
+
+  // TODO: fetch the canonical project from the database
```
thelostone-mc, line 33: Do we have an example I can use as a reference ?

src/indexer/allo/v2/handleEvent.ts
```diff
@@ -25,6 +25,21 @@ function generateRoundRoles(poolId: bigint) {
...
 
+async function getProjectType(db: Database, profileId: Hex, metadata: JSON) {
+
+  if (metadata.hasOwnProperty("canonical")) {
+    // Linked Profile
+    const canonical: CanonicalProject = (metadata as any)["canonical"] as CanonicalProject;
+    const project = await db.getProjectById(canonical.chainId, profileId);
+    if (!project) {
+      throw new Error("Project not found");
+    }
```
gravityblast, line 36: I suggest not to throw an error here. The project might not exist yet if the other chain is slower to index

gravityblast: I would rather set the `type` based on the metadata without checking the db

boudra (changes requested):
I noticed some castings and some issues that will be errors for the linter
I want to suggest this as a more robust way to do the classification (i didn't try the code so it might need modification):

```ts
const linkedProjectMetadata = z.object({
    "canonical": z.object({
           chainId: z.number(),
           registryAddress: z.string()
    })
});

if(linkedProjectMetadata.safeParse(metadata).success) {
   return "linked"
}

return "canonical";
}
```
without the CanonicalProject type.

this way we can start validating the metadatas

src/indexer/allo/v2/handleEvent.ts
```diff
@@ -25,6 +26,21 @@ function generateRoundRoles(poolId: bigint) {
   return { managerRole, adminRole };
 }
 
+function getProjectType(metadata: object) {
+  const linkedProjectMetadata = z.object({
+    canonical: z.object({
+      chainId: z.string(),
```
thelostone-mc, line 32:
```suggestion
      chainId: z.number(),
```

boudra (approved):
⚠️ before you merge please increase the schema version or the schema changes won't be applied: https://github.com/gitcoinco/grants-stack-indexer?tab=readme-ov-file#deployment ⚠️

Issue #442, opened 2024-02-15
index direct grants applications on allo v2


thelostone-mc:
This only adds the Registered event

PR #451, opened 2024-02-16
Index Direct Grants and Applications When Registered

0xKurt:
**Query used  to test**

```
{
  rounds(filter: {tags: {equalTo: "allo-v2"}, id: {equalTo: "28"}}) {
    id
    strategyName
    donationsStartTime
    applicationsStartTime
    tags
    applications {
      id
      status
      metadata
    }
  }
}
```

image

image

fixes https://github.com/gitcoinco/grants-stack-indexer/issues/453
fixes  https://github.com/gitcoinco/grants-stack-indexer/issues/442

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -10,7 +10,7 @@ import os from "node:os";
 type ChainId = number;
 type CoingeckoSupportedChainId = 1 | 10 | 250 | 42161 | 43114;
 
-const CHAIN_DATA_VERSION = "34";
+const CHAIN_DATA_VERSION = "35";
 
 export type Token = {
   code: string;
--- a/src/indexer/allo/v2/handleEvent.ts
+++ b/src/indexer/allo/v2/handleEvent.ts
@@ -1,19 +1,20 @@
 import { EventHandlerArgs } from "chainsauce";
-import { keccak256, encodePacked, pad, decodeAbiParameters, Hex } from "viem";
-import type { Indexer } from "../../indexer.js";
-import {
-  ProjectTable,
-  NewRound,
-  NewApplication,
-  ApplicationTable,
-} from "../../../database/schema.js";
-import { Changeset } from "../../../database/index.js";
+import { Hex, decodeAbiParameters, encodePacked, keccak256, pad } from "viem";
+import { z } from "zod";
 import { parseAddress } from "../../../address.js";
+import { Changeset } from "../../../database/index.js";
+import {
+  NewApplication,
+  NewRound,
+  ProjectTable,
+} from "../../../database/schema.js";
+import type { Indexer } from "../../indexer.js";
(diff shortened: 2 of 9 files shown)


src/indexer/allo/v2/handleEvent.ts
```diff
@@ -452,21 +514,29 @@ export async function handleEvent(
...
-      );
+      switch (round.strategyName) {
+        case "allov2.DirectGrantsSimpleStrategy":
+          values = decodeDGApplicationData(encodedData);
+
+          id = event.params.recipientId;
+          break;
+
+        case "allov2.DonationVotingMerkleDistributionDirectTransferStrategy":
+        default:
```
boudra, line 528: we should throw an error if the strategy is not recognized instead of doing something that might not make sense and cause other problems

src/indexer/allo/v2/handleEvent.ts
```diff
@@ -204,12 +230,18 @@ export async function handleEvent(
...
+            address: strategyAddress,
+          });
+          break;
       }
 
-      let applicationsStartTime = new Date();
-      let applicationsEndTime = new Date();
-      let donationsStartTime = new Date();
-      let donationsEndTime = new Date();
+      let applicationsStartTime: Date | null = new Date();
```
boudra, line 241: it be less error prone if the initial values were null, not sure if there's ever a time that we want to set these dates to the current dates

src/indexer/allo/v2/handleEvent.ts
```diff
@@ -280,15 +335,17 @@ export async function handleEvent(
         applicationsStartTime: isNaN(applicationsStartTime.getTime())
```
boudra, line 335: applicationsStartTime could be null, it's not checked like the other ones

src/indexer/allo/v2/handleEvent.ts
```diff
@@ -255,6 +287,29 @@ export async function handleEvent(
...
+              functionName: "registrationStartTime",
+            }),
+            await readContract({
+              contract,
+              address: strategyAddress,
+              functionName: "registrationEndTime",
+            }),
+          ]);
+
+        // todo: update for Andrea's changes
```
boudra, line 309: what needs to be updated?

src/config.ts
```diff
@@ -10,7 +10,7 @@ import os from "node:os";
 type ChainId = number;
 type CoingeckoSupportedChainId = 1 | 10 | 250 | 42161 | 43114;
 
-const CHAIN_DATA_VERSION = "32";
+const CHAIN_DATA_VERSION = "33";
```
boudra, line 13: i think you need to increase this further

boudra (changes requested):

thelostone-mc (approved):

boudra (approved):

PR #457, opened 2024-02-21
add timezone info to timestamps

boudra:
- pg seems to always store the date in local time instead of UTC, adding timezone information allows us to not lose that information, this is an issue in local development
- saving with the timezone the graphql api returns dates as "2024-02-20T17:00:00+00:00" instead of just "2024-02-20T17:00:00" so they get correctly interpreted as UTC in the front end


bhargavaparoksham (approved):

thelostone-mc (approved):
Ooooo we had to do this on the builder like idiots :P

PR #468, opened 2024-02-28
add canonical project to application

boudra:
- this adds a `canonicalProject` to the application, so that we can get the canonical information without having to query multiple tables

this works for now but there are some considerations:

at the moment the project relationship under applications will error if a project has a linked and a canonical project, try this query, it will fail:

```graphql
{
  applications(
    condition: {chainId: 11155111, roundId: "47"}  ) {
    id
    projectId
    statusSnapshots
    project {
      id
      projectType
    }
  }
```

- technically you can have multiple canonical projects so this canonicalProject relation will not work at some point

- i think the proper solution should be to have a many to many between projects and applications, and have a single project relation as well that takes into account the chain Id of the project

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -10,7 +10,7 @@ import os from "node:os";
 type ChainId = number;
 type CoingeckoSupportedChainId = 1 | 10 | 250 | 42161 | 43114;
 
-const CHAIN_DATA_VERSION = "36";
+const CHAIN_DATA_VERSION = "37";
 
 export type Token = {
   code: string;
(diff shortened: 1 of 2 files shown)


thelostone-mc:
> technically you can have multiple canonical projects so this canonicalProject relation will not work at some point

I don't think this would be the case cause in theory it should have only 1 canonical project even or migrated profiles.
I think your solution would work in every cause we would have

thelostone-mc (approved):

PR #469, opened 2024-02-29
chore: add AlloV1ToV2ProfileMigration

thelostone-mc:
- adds AlloV1ToV2ProfileMigration to be subscribed 
- ProfileMigrated event is indexed
- creating a new table to store the allo-v1 to allo-v2 link 

fix https://github.com/gitcoinco/grants-stack/issues/2846


src/indexer/allo/v2/handleMigratedEvent.ts
```diff
@@ -0,0 +1,62 @@
...
+
+      let changeSet: Changeset[] = [];
+
+      applications.forEach((application) => {
+
+        // Create New Application for alloV2Project
+        const migratedApplication = { ...application };
+        
+        migratedApplication.projectId = alloV2ProfileId;
+        // CALLOUT: This creates a confusion as we may have an application
```
thelostone-mc, line 40: @boudra @0xKurt i don’t think this will be a problem but I wasn’t sure

0xKurt:
Current Problem:
When re-indexing the migrated projects might be indexed before their applications on other chains are indexed, which results in not adding the v1 applications to the migrated v2 project

src/database/index.ts
```diff
@@ -567,6 +578,17 @@ export class Database {
     return application ?? null;
   }
 
+  async getApplicationsByProjectId(chainId: ChainId, projectId: Hex) {
```
thelostone-mc, line 581: I'm assuming this is what we would need to fix @0xKurt ?

src/config.ts
```diff
@@ -588,6 +594,12 @@ const CHAINS: Chain[] = [
         address: "0xDF9BF58Aa1A1B73F0e214d79C652a7dd37a6074e",
         fromBlock: 123566896,
       },
+      // Allo V1 -> V2 Migration
+      // {
+      //   contractName: "AlloV2/AlloV1ToV2ProfileMigration",
+      //   address: "0x",
+      //   fromBlock: 0,
+      // },
```
thelostone-mc, line 602: ^ @boudra This will be updated when we deploy it on live and hence leaving it commented out

src/database/index.ts
```diff
@@ -567,6 +578,17 @@ export class Database {
...
+    const applications = await this.#db
+      .selectFrom("applications")
+      .where("chainId", "=", chainId)
+      .where("projectId", "=", projectId)
+      .selectAll()
+      .execute();
+
+    return applications ?? null;
```
boudra, line 589:
```suggestion
    return applications;
```

this will never be null right? it's always an array

src/database/migrate.ts
```diff
@@ -268,8 +268,18 @@ export async function migrate<T>(db: Kysely<T>, schemaName: string) {
...
+    .addColumn("id", "serial", (col) => col.primaryKey())
+    .addColumn("v1ProjectId", "text")
+    .addColumn("v2ProjectId", "text")
+    .addUniqueConstraint("unique_v1ProjectId", ["v1ProjectId"])
+    .addUniqueConstraint("unique_v2ProjectId", ["v2ProjectId"])
+    .execute();
+
   // https://www.graphile.org/postgraphile/smart-tags/
   // https://www.graphile.org/postgraphile/computed-columns/
+
```
thelostone-mc, line 282:
```suggestion
```

boudra (approved):

PR #478, opened 2024-03-06
add isMatchedAmountDistributed to applications and handle FundsDistributed event

gravityblast:
closes https://github.com/gitcoinco/grants-stack/issues/3081

### DIFF
--- a/src/database/index.ts
+++ b/src/database/index.ts
@@ -594,6 +594,22 @@ export class Database {
     return application ?? null;
   }
 
+  async getApplicationByProjectId(
+    chainId: ChainId,
+    roundId: string,
+    projectId: string
+  ) {
+    const application = await this.#db
+      .selectFrom("applications")
+      .where("chainId", "=", chainId)
+      .where("roundId", "=", roundId)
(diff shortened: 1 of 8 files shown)


src/database/index.ts
```diff
@@ -594,6 +594,22 @@ export class Database {
...
+    const application = await this.#db
+      .selectFrom("applications")
+      .where("chainId", "=", chainId)
+      .where("roundId", "=", roundId)
+      .where("projectId", "=", projectId)
+      .selectAll()
+      .executeTakeFirst();
+
+    return application ?? null;
+  }
```
thelostone-mc, line 611: ^ Just a call out. This won't return the allo-v1 migrated applications given a allo-v2 id

gravityblast: @thelostone-mc  is the migration removing the original application or adding a new application with the new id?

thelostone-mc: migration simply stores a link btwn the old and new id.
we don't actually migrate the old application

thelostone-mc (approved):

PR #479, opened 2024-03-07
update donation voting strategy id to the latest contract version

0xKurt:


src/config.ts
```diff
@@ -10,7 +10,7 @@ import os from "node:os";
 type ChainId = number;
 type CoingeckoSupportedChainId = 1 | 10 | 250 | 42161 | 43114;
 
-const CHAIN_DATA_VERSION = "42";
+const CHAIN_DATA_VERSION = "43";
```
thelostone-mc, line 13:
```suggestion
const CHAIN_DATA_VERSION = "42";
```

PR #523, opened 2024-04-09
add SEI network support

codenamejason:
Adds SEI network support for indexing.


boudra (changes requested):
you have to add chain to INDEXED_CHAINS in fly.staging.toml

boudra (approved):

PR #552, opened 2024-05-15
only enable postgraphile pro if env is present

boudra:
- update some docs and document postgraphile pro
- do not enable postgraphile pro if the env is not set, this is because postgraphile shows a warning that could be confusing
- increase the deploy timeout to 3h, fresh reindexes could take a while


gravityblast (approved):

codenamejason (approved):
Nice, I don't see the warning anymore 🥳

PR #562, opened 2024-05-15
DG: fetch token/match info only if token available

0xKurt:
To enable using custom tokes on direct grants, this PR makes the amount in usd and match token info optional. Since direct grants does not have a matching pool and a match token and we do not display the USD value on the frontend, we default them to 0.

That allows us to insert the application payout without this information, else the payout would be dropped.

### DIFF
--- a/src/indexer/allo/v1/handleEvent.ts
+++ b/src/indexer/allo/v1/handleEvent.ts
@@ -900,25 +900,36 @@ export async function handleEvent(
         return [];
       }
 
-      const amountInUsd = (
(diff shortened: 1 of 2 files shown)


codenamejason (approved):
Looks like it's working 🤗

This is the token caught without price data as exptected -> https://sepolia.etherscan.io/address/0x601Da861335533c38Bc2CAeb90709c5371762559


Screenshot 2024-05-15 at 10 05 22 PM

thelostone-mc (approved):

PR #563, opened 2024-05-15
fix: v2 distribution txn bug

bhargavaparoksham:
fix v2 distribution txn bug

### DIFF
--- a/src/indexer/allo/v2/handleEvent.ts
+++ b/src/indexer/allo/v2/handleEvent.ts
@@ -936,11 +936,11 @@ export async function handleEvent(
       }
 
       const roundId = round.id;
-      const applicationId = event.params.recipientId;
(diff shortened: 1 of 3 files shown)


src/indexer/allo/v2/handleEvent.ts
```diff
@@ -936,11 +936,11 @@ export async function handleEvent(
...
 
       const roundId = round.id;
-      const applicationId = event.params.recipientId;
-      const application = await db.getApplicationById(
+      const anchorAddress = parseAddress(event.params.recipientId);
+      const application = await db.getApplicationByAnchorAddress(
         chainId,
         roundId,
-        applicationId
+        anchorAddress
```
gravityblast, line 943: is there an index in the db for this query to be sure it doesn't take too much?

PR #569, opened 2024-06-11
Grants Stack Indexer MACIQF migration

nijoe1:
- Handled all MACIQF events comprehensively.
- Introduced two new schemas:
     - MACI signups
     - Private donation messages
- Maintained the existing indexer functionality, with one modification:
     - Updated the handling of the RecipientStatusUpdated event types to distinguish between them due to the identical event names.

### DIFF
--- a/src/database/migrate.ts
+++ b/src/database/migrate.ts
@@ -307,6 +307,38 @@ export async function migrate<T>(db: Kysely<T>, schemaName: string) {
     .addUniqueConstraint("unique_v2ProjectId", ["v2ProjectId"])
     .execute();
 
+  await schema
+    .createTable("contributions")
+    .addColumn("id", "text")
+    .addColumn("roundId", "text")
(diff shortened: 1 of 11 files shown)


nijoe1:
On the staging explorer at round/11155111/186, two projects are displayed. However, in the current commit of the indexer, no projects are present. Despite being approved, as verified through GraphQL, these projects are not visible in the explorer. This issue might be due to IPFS failures, as I've observed numerous "failed to parse metadata" errors. I investigated this to ensure that the issue is not solely related to the MACI migration in this PR.

src/database/migrate.ts
```diff
@@ -307,6 +307,38 @@ export async function migrate<T>(db: Kysely<T>, schemaName: string) {
     .addUniqueConstraint("unique_v2ProjectId", ["v2ProjectId"])
     .execute();
 
+  await schema
```
thelostone-mc, line 310: Could we reuse donation and add a data field ?

PR #570, opened 2024-06-12
celo starting blocks + allow null metatada + add totalDistributed

gravityblast:
* Fix Celo starting blocks
* Allow null metadata for projects and rounds created outside grants stack
* Add an index to rounds and projects where metadata is null
* Add `totalDistributed` to rounds


carlbarrdahl (approved):
I'm not very familiar with the indexer code but looks good to me.

src/config.ts
```diff
@@ -19,7 +19,7 @@ type CoingeckoSupportedChainId =
   | 42
   | 42220;
 
-const CHAIN_DATA_VERSION = "67";
+const CHAIN_DATA_VERSION = "71";
```
gravityblast, line 22: the current version in `main` is 70

PR #657, opened 2024-09-04
index single chain

0xKurt:
fixes PAR-329

### DIFF
--- a/src/database/index.ts
+++ b/src/database/index.ts
@@ -994,35 +994,42 @@ export class Database {
         .where("chainId", "=", chainId)
         .execute();
 
-      this.#logger.info("Deleting applications payouts");
+      this.#logger.info("Deleting applications");
       await trx
         .withSchema(this.chainDataSchemaName)
         .deleteFrom("applications")
         .where("chainId", "=", chainId)
         .execute();
 
-      this.#logger.info("Deleting applications payouts");
+      this.#logger.info("Deleting applications donations");
       await trx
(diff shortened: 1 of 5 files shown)


src/http/app.ts
```diff
@@ -100,6 +100,33 @@ export const createHttpApi = (config: HttpApiConfig): HttpApi => {
...
+  app.get("/index", (req, res) => {
+    const chainId = req.query.chainId;
+
+    if (!chainId) {
+      return res.status(400).send("chainId is required");
+    }
+
+    const reindex = async () => {
+      // todo, why doesn't this work? - it's not deleting the indexed data.
+      await config.db.deleteChainData(Number(chainId));
```
0xKurt, line 117: @thelostone-mc

src/database/index.ts
```diff
@@ -791,4 +791,72 @@ export class Database {
 
     return result ?? null;
   }
+
+  async deleteChainData(chainId: ChainId) {
+    this.#logger.info("Deleting chain data for chainId:", chainId);
+
+    await this.#db.transaction().execute(async (trx) => {
```
thelostone-mc, line 798: - rename to tx ?
- Also we should likely do `this.#db.withSchema("schemaName").trans...` cause else this will break once we add the IPFS schema

0xKurt: @thelostone-mc will update this after your PR was merged, ao  i can rebase and have the new schema names available

thelostone-mc (approved):

PR #665, opened 2024-09-09
improve locking

thelostone-mc:
- Allow being able to write to chain schema even if the machine hasn't gotten lock for schemas  
- ipfs and price attempt to forcible acquire lock (this has to be tested)

### DIFF
--- a/src/database/index.ts
+++ b/src/database/index.ts
@@ -102,6 +102,12 @@ export class Database {
       }, 0);
     };
 
+    // Helper function to forcibly acquire a lock for a specific schema
+    const forciblyAcquireLockForSchema = async (lockId: number) => {
+      // This will block until the lock is acquired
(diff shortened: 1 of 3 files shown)


src/database/index.ts
```diff
@@ -121,26 +127,40 @@ export class Database {
...
-          },
-          client,
-        };
-      }
+      if (chainDataLockAcquired) acquiredLocks.push(chainDataLockId);
+
+      // const ipfsDataLockAcquired = await acquireLockForSchema(ipfsDataLockId);
+      // if (ipfsDataLockAcquired) acquiredLocks.push(ipfsDataLockId);
+      // const priceDataLockAcquired = await acquireLockForSchema(priceDataLockId);
+      // if (priceDataLockAcquired) acquiredLocks.push(priceDataLockId);
```
thelostone-mc, line 140: ^ this is soft attempt lock

PR #673, opened 2024-09-18
chore: add custom fields

thelostone-mc:
closes PAR-396

Deploy

@umar / @ed Got a PR up which should give you

- blocknumber when distribution was uploaded **blocknumber**
- blocktimestamp when distribution was uploaded **blocktimestamp**
- closest usdPrice of matching token when distribution was uploaded **usdPrice**
- time at which the usd price was recorded **usdPriceTimestampAt**


hussedev (approved):
LGTM

PR #674, opened 2024-09-23
index Gitcoin Attestation Network

thelostone-mc:
Close PAR-407

This adds 2 new tables
- Attestations: which indexes the summary and metadata of the actual cart / checkout
- AttesationTxn : which indexes all the attesation txn and which chain the attesation lives on

There is also a fk to the donations table which lets you list donations for a given a given AttesationTxn and also check if a donation has an associated attestation uid

```
{
  attestationTxns {
    chainId
    attestation {
      uid
    }
    donations {
      id
    }
  }
}
```

```
{
  donations {
    id
    attestationTxnsByTxnHashAndChainId {
      attestationUid
    }
  }
}
```

Deploy

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -21,7 +21,7 @@ type CoingeckoSupportedChainId =
   | 42220
   | 1088;
 
-const CHAIN_DATA_VERSION = "83";
+const CHAIN_DATA_VERSION = "86";
 const IPFS_DATA_VERSION = "1";
 const PRICE_DATA_VERSION = "1";
 
--- a/src/config.ts
+++ b/src/config.ts
@@ -368,6 +368,12 @@ const CHAINS: Chain[] = [
         address: "0xCd5AbD09ee34BA604795F7f69413caf20ee0Ab60",
         fromBlock: 5100681,
       },
+      // Gitcoin Attestation Network
+      {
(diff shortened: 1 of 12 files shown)


src/indexer/gitcoin-attesation-network/handleEvent.ts
```diff
@@ -0,0 +1,112 @@
...
+        return [];
+      }    
+
+      let donations = [];
+      for (let i = 0; i < data.length; i++) {
+        const metadata = data[i];
+        donations = await db.getDonationsByTxHash(metadata.chainId, metadata.txnHash);
+        // should we just link the donations table to the attestation?
+        // this means we would only need the txnHash in the attestation metadata
+        // all the other data can be fetched via getDonationsByTxHash
```
thelostone-mc, line 84: @0xKurt @hussedev @nijoe1

nijoe1: Makes sense. But if others want to use the metadata differently, it might make things harder for them. If that’s not a big deal, we can just include the chainId and transactionHash. Otherwise, we should add everything.

hussedev (approved):
LGTM!

